PR #1, opened 2015-09-23
replace HEAD placeholder (rather than just title)

ronco:
Companion PR for https://github.com/tildeio/ember-cli-fastboot/pull/66.  Allow fastboot to provide an entire head, rather than just title.


ronco:
This was accomplished with #15

PR #2, opened 2015-10-08
Use latest visit API

chancancode:
Update fastboot server for https://github.com/emberjs/ember.js/pull/12394

Require coordination from https://github.com/tildeio/ember-cli-fastboot/pull/71

- [x] Take advantage of the new API
- [x] FastFail™
- [x] Don't mutate `autoBoot`
- [x] Remove `FastBoot` global
- [x] Better integration with Ember CLI
- [x] Move most of the work out of initializers
- [x] Test that it works
- [ ] Write tests

Open question:

* What are the rules regarding passing complex objects across the sandbox boundary? It would not be hard to avoid that if we have to.

cc @tomdale @stefanpenner

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -8,23 +8,20 @@ var najax = require('najax');
 var debug   = require('debug')('ember-cli-fastboot:ember-app');
 var emberDebug = require('debug')('ember-cli-fastboot:ember');
 
+var HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
+
 function EmberApp(options) {
-  this.appFile = options.appFile;
-  this.vendorFile = options.vendorFile;
+  var appFilePath = options.appFile;
+  var vendorFilePath = options.vendorFile;
 
-  debug("app created; app=%s; vendor=%s", this.appFile, this.vendorFile);
-
-  // Promise that represents the completion of app boot.
-  var appBoot = RSVP.defer();
+  debug("app created; app=%s; vendor=%s", appFilePath, vendorFilePath);
 
   // Create the sandbox, giving it the resolver to resolve once the app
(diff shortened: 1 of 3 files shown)


lib/ember-app.js
```diff
@@ -32,16 +29,38 @@ function EmberApp(options) {
   sandbox.run(appFile);
   debug("app file evaluated");
 
-  this.waitForBoot = function() {
-    return appBoot.promise;
-  };
+  var AppFactory = sandbox.require('~fast-boot/app-factory');
```
rwjblue: Pardon my ignorance, but what does `~` do here?

chancancode: I copied it from Stef :P I think it is just to make it less likely to collide with user codez; on the other hand, I already relied on a globa called `FastBoot` so...

chancancode: This is injected from the addon build step

rwjblue:
LGTM

@tomdale - Can you review?

lib/ember-app.js
```diff
@@ -8,23 +8,20 @@ var najax = require('najax');
...
-  // Promise that represents the completion of app boot.
-  var appBoot = RSVP.defer();
+  debug("app created; app=%s; vendor=%s", appFilePath, vendorFilePath);
 
   // Create the sandbox, giving it the resolver to resolve once the app
   // has booted.
-  var sandbox = createSandbox(appBoot.resolve, {
-    najax: najax
-  });
+  var sandbox = createSandbox({ najax: najax });
```
tomdale: We should think about exposing a more flexible primitive for this so addons can do similar things.

tomdale:
Awesome work, @chancancode!

tomdale:
Published this as `v0.1.0`.

PR #5, opened 2015-10-24
Add clearTimeout to the sandbox

### LINKED ISSUE #4: Sandbox is missing clearTimeout
clearTimeout is used by backburner, it should be exposed to the sandbox like is done with setTimeout.

mattclaycomb:
This fixes #4


tomdale:
Thank you

PR #8, opened 2016-01-08
Misc work on the ember-fastboot-server project

jmurphyau:
- Added static assert serving
- Allowed multiple js files to be passed
- Refactored sandboxes and added the node-vm sandbox (for node v0.12.0+) - this is recommended by the contextify author on the repo's main page.
- Added the host option - in newer versions of node it defaults to IPv6 - this option forces 0.0.0.0 but allows users to specify something different

### DIFF
--- a/bin/ember-fastboot
+++ b/bin/ember-fastboot
@@ -7,11 +7,19 @@ process.title = 'ember-fastboot-server';
 var FastBootServer = require('../lib/server');
 var express = require('express');
 var parseArgs = require('minimist');
+var path = require('path');
+var glob = require('glob');
+var assert = require('assert');
 
 var options = parseArgs(process.argv.slice(2));
+var distFile = options['dist-file'];
 var appFile = options['app-file'];
 var vendorFile = options['vendor-file'];
 var htmlFile = options['html-file'];
+
+if (distFile) {
+  findFromDist();
+}
 
 if (!appFile) {
   console.error("You must specify the path to an application file with the --app-file option.");
@@ -79,3 +87,30 @@ server.app.boot().then(function() {
 
   process.exit(1);
 });
+
+function findFromDist() {
+  appFile = findAppFile();
+  vendorFile = findVendorFile();
+  htmlFile = findHTMLFile();
+}
+
+function findAppFile() {
+  return findFile("app", path.join(distFile, "assets", "!(vendor*).js"));
+}
+
+function findVendorFile() {
+  return findFile("vendor", path.join(distFile, "assets", "vendor*.js"));
+}
+
+function findHTMLFile() {
+  return findFile('html', path.join(distFile, 'index*.html'));
+}
+
+function findFile(name, globPath) {
+  var glob = require('glob');
+  var files = glob.sync(globPath);
+
+  assert(files.length === 1, "Found " + files.length + " " + name + " files (expected 1) when globbing '" + globPath + "'.");
+
+  return files[0];
+}


README.md
```diff
@@ -49,3 +52,42 @@ var listener = app.listen(options.port, function() {
...
+    ....
+```
+
+### Debugging
+
+To debug your fastboot app you can use the `node-debug` to start the `ember-fastboot` command line tool.
+
+Some tips:
+
+* If you have trouble with modules not found, try use the `--cli` options - this removes the execution of anyway
```
tomdale: Wrong word here?

README.md
```diff
@@ -14,14 +14,19 @@ prerequisite for developing FastBoot apps.
...
-    --app-file path/to/app-file.js
-    --vendor-file path/to/vendor-file.js
-    --html-file path/to/index.html
+    --host 0.0.0.0 \
+    --port 80 \
+    --html-file path/to/index.html \
+    path/to/vendor-file.js \
+    path/to/app-file.js \
+    path/to/other-file-1.js \
+    path/to/other-file-2.js
```
tomdale: What is the use case for this?

tomdale: Also how does the `--dist-file` option interact with the list of files? Seems like they are in conflict.

jmurphyau: I got the idea from when you mentioned you're working on testing stuff and thinking about how I would go about executing tests using the fastboot command line.. For example if you had the fastboot specific test code and it needs to run after all your JS or before it all - or in-between some groups of files.. 

```
ember-fastboot \
    --host 0.0.0.0 \
    --port 80 \
    --html-file dist/tests/index.html \
    path/to/vendor-file.js \
    path/to/app-file.js \
    path/to/fastboot-test-support.js
```

Also the thought was there that the ability to split the files into more than just two (vendor/app) could allow for more flexible or smarter live code reloading (whenever/if ever that comes) as well as potentially reusing the files that are compiled with regular `ember build` command - e.g. if whatever is added/changed in your app by using the command `ember fastboot:build` could instead be split out into dedicated files - then there wouldn't be a need to recreated vendor/app

Also partly influenced by not wanting to type --app-file and --vendor-file

README.md
jmurphyau: yep

jmurphyau:
`dist-file` `app-file` and `vendor-file` are keep just to make life easier by not breaking the way it works now - the idea was to move from those options to just providing files and then remove those options after some period of time after transitioning

README.md
tomdale: My concern is that it pushes complexity to the end user, who has to know the precise order to invoke these things. I think moving towards `--dist-file=` or just pointing it at the directory as an argument is better. Forcing users to think about the specific order files are evaluated in seems unnecessary.

tomdale:
@jmurphyau Do you think you could incorporate https://github.com/samccone/engine-deps into this PR so that contextify only gets installed for Node 0.10 and lower? Currently, it adds significant time to the install process, and building native dependencies is fraught with peril.

README.md
jmurphyau: Yeah that makes sense and I agree - I'll have another think about this

jmurphyau:
yep no worries will do

tomdale:
@jmurphyau After thinking about this a bit more, my inclination is that we should just standardize on a directory layout for FastBoot builds and have you point the server at a directory with everything in the correct place. What do you think? I'm going to be doing some work that will rely on reading a manifest/configuration file, and the number of increasing configuration options is getting ridiculous. ;)

jmurphyau:
@tomdale Yep sounds like a good idea.. If we can just point to one place it will make life a lot easier

tomdale:
@jmurphyau Any update on this?

jmurphyau:
I should have this back up to date within the next day or two

tomdale:
@jmurphyau Awesome, excited to see it. Thanks!

PR #9, opened 2016-01-22
Only Use Contextify When Required

jmurphyau:
I didn't realise my other PR was closed.. It seems they close themselves when the branch is exactly the same as the branch you're doing the PR against..

This PR adds the host option and uses/installs contextify only when it's needed (< node 0.10.x)

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -1,7 +1,8 @@
 var fs = require('fs');
 var path = require('path');
 
-var Contextify = require('contextify');
+var createSandbox = require('./sandbox').createSandbox;
+var runInSandbox = require('./sandbox').runInSandbox;
 var SimpleDOM = require('simple-dom');
 var RSVP    = require('rsvp');
 var chalk = require('chalk');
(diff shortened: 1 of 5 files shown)


tomdale:
Thanks @jmurphyau, can you move the host change into a separate PR? There are still a few commented out lines that should be cleaned up as well. Lastly, please use the style `isLegacyVM` instead of `isLegacyVM`. I will try to add a style guide so that isn't ambiguous. Thanks!

jmurphyau:
@tomdale should look a bit better now

jmurphyau:
Also, just an FYI - that plugin you mentioned about installing dependancies based on node version.. What it did was install a version of npm (which could be (and in my case was) a different version to the users current npm)  as a dependancy and then programatically install the dependancies.. The tried this and I had two issues with it:

1. It downloaded NPM as a dependancy.. You mentioned the length of time for building the contextify plugin - it almost seemed like downloading NPM took the same duration or longer.
2. NPM has been recently updated to show a progress bar sort of things rather than the previous style (I think it output a list of things that were happening when they were happening?) - the plugin used an NPM that produced the older style output.. It threw me off for a split second before realising what it was - especially since I was switching between node and NPM versions to test it out..

So instead of using that I've gone with a script that checks the `isLegacyVM` return result and produces an exit code based on that together with post install config that will install contextify if required

tomdale:
Awesome! Thank you for all the hard work. I'm very excited about this change.

tomdale:
@jmurphyau I had to revert this because it causes npm installs of this package to fail if it's not a legacy VM.

jmurphyau:
really? hmm thats odd.. I use the latest node/npm - which fall into the category you mention - and it appeared to work fine for me.. Do you have any example of the error that appears

PR #12, opened 2016-02-03
Mixed issue with NPM install failing

jmurphyau:
This puts back the contextify conditional install - but correctly .. The previous attempt had a major issue with the install script.

'is-not-legacy-vm' could do with a better name I think


tomdale:
@jmurphyau I had to revert the original commit, I don't think this PR includes adding it back. Would you mind doing that and rebasing it? Looks good otherwise.

tomdale:
Closed in favor of #36

PR #17, opened 2016-02-15
Pass express.js' request/response to cookie service

marcoow:
This passes express.js' request and response to the new cookie service introduced by tildeio/ember-cli-fastboot#120 so that that service can implement cookie access by wrapping the respective methods from the request/response object or by modifying `document.cookie` when in browser mode.

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -55,9 +55,10 @@ EmberApp.prototype.boot = function() {
 };
 
 EmberApp.prototype.visit = function(url, options) {
+  console.log('visit options:', options);
   var doc = new SimpleDOM.Document();
   var rootElement = doc.body;
-  var options = { isBrowser: false, document: doc, rootElement: rootElement, cookies: options.cookies };
+  var options = { isBrowser: false, document: doc, rootElement: rootElement, request: options.request, response: options.response };
 
   return this._app.visit(url, options).then(function(instance) {
     var head;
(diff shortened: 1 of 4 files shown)


marcoow:
Of course an alternative approach would be to just serialize all the cookies from the request into the `cookie` property of the `SimpleDOM` document before calling `visit` on the app instance and then before responding convert all (potentially changed) cookies from the `cookie` property of the `SimpleDOM` document into actual cookies on the express.js response again so that for the Ember app the API to interact with cookies would always be just `document.cookie` like in browser envs.

Just realize that approach might actually be a bit nicer…

lib/ember-app.js
```diff
@@ -50,30 +50,41 @@ function EmberApp(options) {
...
-      };
-    } finally {
-      instance.destroy();
-    }
+  return this.boot(options).then(function(instance) {
+    var cookiesService = instance.lookup('service:cookies');
+    cookiesService.setProperties({
+      request: request,
+      response: response
+    });
```
marcoow: This only runs after all the application's instance initializers have completed which doesn't really work of course as the cookie service won't be able then to read/write cookies then if used in an instance initializer.

tomdale:
@marcoow I just submitted #19 which is this PR + the changes we discussed on Slack.

rwjblue:
I'm going to close this for now, in favor of #19.

marcoow:
👍🏻

PR #19, opened 2016-02-23
Handle cookies

tomdale:
This slightly changes #17 to use the FastBoot service, and does the cookie parsing on the server side.

/cc @rwjblue @marcoow

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -9,6 +9,7 @@ var najax = require('najax');
 var debug   = require('debug')('ember-cli-fastboot:ember-app');
 var emberDebug = require('debug')('ember-cli-fastboot:ember');
 var sourceMapSupport = require('./install-source-map-support');
+var cookie = require('cookie');
 
 var HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
 
@@ -60,18 +61,26 @@ EmberApp.prototype.boot = function(options) {
   });
 };
 
-EmberApp.prototype.visit = function(request, response, options) {
+EmberApp.prototype.visit = function(request, response) {
   var path = request.path;
   var doc = new SimpleDOM.Document();
   var rootElement = doc.body;
   var options = { isBrowser: false, document: doc, rootElement: rootElement };
 
   return this.boot(options).then(function(instance) {
-    var cookiesService = instance.lookup('service:cookies');
-    cookiesService.setProperties({
-      request: request,
-      response: response
+    var fastbootService = instance.lookup('service:fastboot');
+
+    var cookies = request.get('cookie');
+    if (cookies) {
+      cookies = cookie.parse(cookies);
+    }
+
+    fastbootService.setProperties({
+      _request: request,
+      _response: response,
+      cookies: cookies
     });
+
 
     return instance.visit(path, options).then(function(instance) {
       var head;
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -54,31 +54,41 @@ function EmberApp(options) {
   this._app = AppFactory['default']();
 }
 
-EmberApp.prototype.boot = function() {
-  return this._app.boot();
+EmberApp.prototype.boot = function(options) {
+  return this._app.boot().then(function(app) {
+    return app.buildInstance().boot(options);
+  });
 };
 
-EmberApp.prototype.visit = function(url, options) {
-  console.log('visit options:', options);
+EmberApp.prototype.visit = function(request, response, options) {
+  var path = request.path;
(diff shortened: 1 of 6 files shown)


lib/ember-app.js
```diff
@@ -54,30 +55,49 @@ function EmberApp(options) {
...
+
+    var cookies = request.get('cookie');
+    if (cookies) {
+      cookies = cookie.parse(cookies);
     }
+
+    fastbootService.setProperties({
+      _request: request,
+      _response: response,
+      cookies: cookies
```
rwjblue: Should `cookies` also be private for now (i.e. `_cookies`)?

rwjblue:
LGTM, but I'd like to see this kept as private for now while we settle in on things (same as `_request` / `_response`)....

lib/ember-app.js
```diff
@@ -54,30 +55,49 @@ function EmberApp(options) {
...
-      instance.destroy();
+  return this.boot(options).then(function(instance) {
+    var fastbootService = instance.lookup('service:fastboot');
+
+    var cookies = request.get('cookie');
+    if (cookies) {
+      cookies = cookie.parse(cookies);
     }
+
+    fastbootService.setProperties({
```
marcoow: This still has the problem that cookies are only accessible after the application instance has booted (e.g. you cannot use cookies in initializers/instance initializers). Not sure whether using cookies in (instance) initializer should be considered an anti pattern but I guess it might be surprising to people that they can only access cookies after all instance initializers have run?

tomdale:
@rwjblue I think it's OK to make `cookies` private since:

1. 100% of people who have authenticated apps need it.
2. The surface area is quite small: it's just a POJO with cookies as key/values.

Given that we're pre-1.0, I _think_ we could get away with changing it if we had to. But given how simple it is, I think it's probably OK.

lib/ember-app.js
tomdale: @marcoow Yes, this is definitely a tradeoff of this implementation. I went spelunking in the Ember `Application` and `ApplicationInstance` code and it seems there is not an elegant way of setting cookies prior to running instance initializers.

That said, I just re-read the code now and I think I misread last night. I thought that registry creation and running of instance initializers all happened in [_bootSync](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application-instance.js#L140-L169). But actually it looks like `__registry__` exists on the application instance before the boot process begins.

I _think_ that means that we could register an object in the registry before the instance initializers are kicked off. It involves reaching pretty deep into private API, so we should think about what we're trying to express here and see if there's an elegant way of exposing it as public API.

marcoow: I think restoring authentication state (or user settings etc.) from cookies in (instance) initializers is probably still a pretty common thing to do. In fact, Ember Simple Auth did just that until version 1.0 as well.

I think the `ApplicationInstance` returned from `buildInstance` (https://github.com/ember-fastboot/ember-fastboot-server/pull/19/files#diff-ba82abedace362f3becf198dd49dacb4R60) already has the `register` method so that it should be possible to register the request and response as `-request:main` and `-response:main` and inject these into the `fastboot` service then?

tomdale: @marcoow Yes, I think you're right. I'm in the middle of updating the PR now to attempt this.

rwjblue:
@tomdale - My personal preference is to err on the side of private and make it public once we are certain. I am happy to defer to you on this.

tomdale:
@rwjblue @marcoow I'm pretty happy with this latest update. Please let me know what you think.

rwjblue:
Seems good to me. 

BTW, I'm starting to feel the desire for some tests in this repo....

marcoow:
Looks great! I think in the long term there would have to be a public interface to `request` and `response` but it's probably better to keep them (or actually the complete `FastBootInfo`) private for now and see how everything goes from here.

tomdale:
@rwjblue FWIW all of these new features I'm working on come with tests in the `ember-cli-fastboot` repo. There is an extremely tight coupling going on between the two and I don't have a strong sense of how to test FastBootServer without a compiled Ember app, which requires `ember-cli-fastboot`, which requires `ember-fasboot-server`, which requires…

That said, yeah, I agree it's problematic. We should try to think of something. We can at minimum probably write a stub Ember app generator that generates the minimum scaffolding required to trick the server into thinking it's running an Ember app.

marcoow:
@tomdale: I published (a very early version of) an addon that abstracts cookie access so it works both in the browser as well as with Fastboot here: https://github.com/simplabs/ember-cookies

tomdale:
@marcoow Awesome!

PR #20, opened 2016-02-23
Allow for document to be provided programmatically

jasonmit:
This adds the ability for an upstream middleware to provide the document through `res.locals.fastbootHTML` versus loading it from disk.  This enables setting the initial title, meta tags, pulling it from a redis server, preloading the document with translations based on the requested url, etc. etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,44 @@ var listener = app.listen(process.env.PORT || 3000, function() {
   console.log('FastBoot running at http://' + host + ":" + port);
 });
 ```
+
+#### Providing Index Document Programmatically
+
+You may not store `index.html` with the fastboot build, perhaps it's
+stored on an external server, or want to preload the document with additional
+information such as i18n translations.  To do this, first, omit `htmlFile` from the options hash
+that is provided to the `FastBootServer` constructor.
+
+Next, register a middleware _before_ `server.middleware`, retrieve your index file,
+or read it from disk and manipulate the content.  Finally, set the final value to `res.locals.fastbootHTML`.
+
+```js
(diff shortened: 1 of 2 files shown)


rwjblue:
I'm generally :+1: to allowing this to be provided programmatically. Can you paste a snippet on how this might be used in the `README.md` or something?

jasonmit:
@rwjblue all set

tomdale:
I wonder if we shouldn't make it `res.locals.fastboot.html` rather than `res.locals.fastbootHTML`. Presumably in the future we'll want to add more and it's easier to put them all in one object rather than namespace each property.

README.md
```diff
@@ -42,3 +42,44 @@ var listener = app.listen(process.env.PORT || 3000, function() {
   console.log('FastBoot running at http://' + host + ":" + port);
 });
 ```
+
+#### Providing Index Document Programmatically
+
+You may not store `index.html` with the fastboot build, perhaps it's
```
tomdale: This reads strangely to me. Seems like there's a missing "want to"?

jasonmit:
> I wonder if we shouldn't make it res.locals.fastboot.html rather than res.locals.fastbootHTML.

That sounds like a better idea :+1:

README.md
jasonmit: :+1:

jasonmit:
Actually I thought more about `res.locals.fastboot.html` and it will require users to create the object if one isn't created before setting `html`, or any future properties.  Do you think we should extend the response object and expose a setter/getter for nicer developer ergonomics?

jasonmit:
No longer needed since you can pass `html` into the `options` hash.

PR #21, opened 2016-02-24
Retain query params when routing

tomdale:
This commit fixes a bug where query params were inadvertently dropped
from the incoming request when routing in Ember.

See https://github.com/tildeio/ember-cli-fastboot/issues/53


tomdale:
#yolomerge

PR #24, opened 2016-02-27
Update najax release

danmcclain:
0.4.0 has the functions necessary for ember-data to work in fastboot


tomdale:
@danmcclain I'd like to eventually remove hard-coded `najax` support and just have Ember Data require it using the `fastbootDependencies` option. Should be straightforward now that ED is an npm package.

PR #25, opened 2016-02-29
Add host to `FastBootInfo`

danmcclain:


lib/fastboot-info.js
```diff
@@ -30,6 +31,10 @@ FastBootInfo.prototype.extractCookies = function(request) {
   return {};
 };
 
+FastBootInfo.prototype.extractHost = function(request) {
+  return `${request.protocol}://${request.headers.host}`;
```
tomdale: Shouldn't this be `request.hostname` instead of `request.headers.host`?

From the [Express docs](http://expressjs.com/en/api.html#req.hostname):

> *req.hostname*
> Contains the hostname derived from the Host HTTP header.

danmcclain: I'm using the host header because it also contains the port, and instead of reassembling that data from the parts parsed from the header, I'm just using the compiled header. If there were subdomains, we'd have to reconcatenate them before assembling the host

danmcclain:
@tomdale Updated to take into account the `hostWhitelist` that will be added to the package.json

lib/fastboot-info.js
```diff
@@ -30,6 +31,20 @@ FastBootInfo.prototype.extractCookies = function(request) {
...
 };
 
+FastBootInfo.prototype.host = function() {
+  if (!this.hostWhitelist) {
+    throw new Error('You must provide a hostWhitelist to retrieve the host');
+  }
+
+  var host = this.request.headers.host;
+
+  if (this.hostWhitelist.indexOf(host) === -1) {
```
tomdale: This should probably support regex or wildcards, so you can say `"*.example.com"` instead of enumerating all of the possible combinations.

danmcclain: Was on the fence of that, went explicit, but will update this in a bit :+1:

danmcclain:
@tomdale Updated for regex entries in whitelist

tomdale:
Tests need to pass on Node 0.10. :dancer:

danmcclain:
:dancer: Fixed @tomdale

PR #26, opened 2016-03-01
Application#boot was renamed to buildApp

thorsteinsson:
This resulted in a TypeError in bin/ember-fastboot. The change was made in deabcea.


tomdale:
Good catch, thank you. I will add some tests to the binary so this doesn't happen again.

PR #27, opened 2016-03-02
Basic tests + improved error messages

tomdale:
More to come, but this sets up a basic testing harness and should allow us to accept pull requests with more confidence, and without the synchronization song and dance usually required with `ember-cli-fastboot`.

/cc @danmcclain @chancancode @rwjblue


danmcclain:
:+1: I can help with tests for headers. I'll make sure that host has tests as well

PR #28, opened 2016-03-03
Add static asset serving + CLI test infrastructure

tomdale:
This adds the ability to serve static assets directly from the FastBoot server if you are not using a CDN. This commit also adds testing infrastructure and tests for the CLI part of the package.


gopeter:
Could you please update the docs on how to serve static assets when using it along with `express`? 

This would be very great :)

tomdale:
@gopeter Good idea. :) Do you want to take a stab at this and submit a PR? I can help with any questions you have.

gopeter:
I'm not sure if I worked it out correctly, but this is how I solved it for me since I couldn't find any option/configuration for `FastBootServer` to serve static assets:
````
app.get('/', server.middleware());
app.use('/assets', express.static('dist/assets'));
````
If this is the right way, I'll submit a PR :)

tomdale:
@gopeter That looks right to me, I think that's what we're doing in the CLI version.

gopeter:
OK, got it to work now (#44) with all types of assets :) Maybe this could be added to the CLI tests.

PR #29, opened 2016-03-05
Actually pass the hostWhitelist into FastBootInfo

danmcclain:
@tomdale Not entirely sure the best way to test that the config is passed in (other than running it)


danmcclain:
This will end up getting tested when I finish implementing it into the service in `ember-cli-fastboot` (which is how I found it)

PR #36, opened 2016-04-04
Remove hard contextify dependency

tomdale:
This is based on @jmurphyau’s previous work, but updated to use subclasses. Hopefully in the future we can make this pluggable.

This detects if the current version of Node being used supports newer `vm` features, and if so, skips installing `contextify`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Production server for running Ember applications using FastBoot",
   "main": "./lib/server",
   "scripts": {
+    "postinstall": "scripts/is-not-legacy-vm || npm install contextify@^0.1.11",
     "test": "mocha"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   "homepage": "https://github.com/ember-fastboot/ember-fastboot-server#readme",
   "dependencies": {
     "chalk": "^0.5.1",
-    "contextify": "^0.1.11",
     "cookie": "^0.2.3",
     "debug": "^2.1.0",
     "express": "^4.13.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Production server for running Ember applications using FastBoot",
   "main": "./lib/server",
   "scripts": {
+    "postinstall": "scripts/is-not-legacy-vm || npm install contextify@^0.1.11",
```
ronco: This isn't working as I would expect.  `is-not-legacy-vm` is reporting as new in every version of node I've tried (back to 0.12).  So it never falls through to the conditional install.  `contextify` is installed, but that's because it's still in the package.json.

tomdale: @ronco The legacy VM behavior is only necessary on 0.10, so what you're describing sounds correct. Good catch on the lingering `contextify` dep in `package.json`; I've removed it.

test/helpers/cli-server.js
```diff
@@ -32,12 +32,12 @@ Server.prototype.start = function() {
     });
 
     server.on('exit', function(statusCode, signal) {
-      if (signal === 'SIGTERM') {
+      if (signal === 'SIGTERM' || statusCode === 143) {
```
rwjblue: What is this change for?  How is it related to the contextify changes?

rwjblue:
I left a tiny question, but this looks great to me...

package.json
ronco: Ah, gotcha, I thought it was a newer feature than that. :+1:

rwjblue: Nope. We are intentionally supporting the oldest possible node engine (which looses official support on 2016-10-01) in order to support AWS lambda. 

We do need a larger plan for dealing with node version support though.

test/helpers/cli-server.js
tomdale: @rwjblue This change is just because on Node 0.10 it returns a status code of 143 instead of the string `SIGTERM`.

PR #37, opened 2016-04-06
[MINOR]: Fix "main" in package.json

kratiahuja:
While looking at ember-fastboot-server for sandbox reuse, I realized the entry point of this package was incorrect. There is no `./lib/server` anymore. So I changed it to point to the `lib` folder.

I did not add any tests since this was more of a cleanup.


kratiahuja:
My bad on not reading the full package correctly.

PR #38, opened 2016-04-11
Provide FastBootRequest object to FastBootInfo

danmcclain:
Capture query params and path off the request

I think we may want to remove `FastBootInfo.request` and replace it with an object that we build ourselves. In that effort, `FastBootInfo.{cookies,headers}` would belong in that `request` object. This would obviously change the API here and should likely be mirrored in `ember-cli-fastboot` (with proper deprecation warnings)

~~This PR is made without doing the above, storing the query params and path on the `FastBootInfo` root~~

@tomdale Thoughts on this?

### DIFF
--- a/test/fastboot-headers-test.js
+++ b/test/fastboot-headers-test.js
@@ -1,0 +1,68 @@
+var expect = require('chai').expect;
+var path = require('path');
+var FastBootHeaders = require('../lib/fastboot-headers.js');
+
+describe('FastBootHeaders', function() {
+  it('returns an array of header values from getAll, regardless of header name casing', function() {
+    var headers = {
+      // Express concatenates repeated keys with ', '
+      // and also lowercases the keys
+      'x-test-header': 'value1, value2'
+    };
+    var headers = new FastBootHeaders(headers);
+
+    expect(headers.getAll('X-Test-Header')).to.deep.equal(['value1', 'value2']);
+    expect(headers.getAll('x-test-header')).to.deep.equal(['value1', 'value2']);
+  });
+
+  it('returns an emtpy array when a header is not present', function() {
+    var headers = {
+      // Express concatenates repeated keys with ', '
+      // and also lowercases the keys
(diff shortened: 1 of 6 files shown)


tomdale:
@danmcclain Yes, this sounds good. There has been some rumbling around switching from Express to Hapi, providing our own `request` object with explicitly copied properties would help mitigate any accidental breakage if we do decide to that. (Or if someone uses the FastBoot middleware with a non-Express server.)

danmcclain:
@tomdale Updated. This implements `FastBootRequest` now

test/fastboot-request-test.js
```diff
@@ -0,0 +1,165 @@
...
+      }
+    };
+    var hostWhitelist = ["example.com", "localhost:4200"]
+    var fastbootRequest = new FastBootRequest(request, hostWhitelist);
+
+    var host =  fastbootRequest.host();
+    expect(host).to.equal("localhost:4200");
+  });
+
+  it("returns the host matches a regex in the hostWhitelist", function() {
```
tomdale: Typo here: should probably be "returns the host if it matches a regex"

tomdale:
Now that we have our own request object, I'm wondering if we should expose something à la the `get()` method that normalizes header names, since their case insensitive and having to `.toLowerCase()` it yourself everywhere is pretty annoying.

danmcclain:
@tomdale Added FastBootHeaders

test/fastboot-headers-test.js
```diff
@@ -0,0 +1,68 @@
...
+      // and also lowercases the keys
+      "x-test-header": "value1, value2"
+    };
+    var headers = new FastBootHeaders(headers);
+
+    expect(headers.get('X-Test-Header')).to.equal('value1');
+    expect(headers.get('x-test-header')).to.equal('value1');
+  });
+
+  it('returns the null when using get when a header is not present', function() {
```
pwfisher: typo, extra "the"

PR #39, opened 2016-04-11
Adds hot reloading/zero downtime deploy

tomdale:
This commit adds support for reloading a running FastBoot server.

In the programmatic case, you can call `server.reload()` to have it
destroy the old app and re-read the new version off the disk.

If running from the command line, you can send SIGUSR1 to the process
and it will reload the app from disk.


package.json
```diff
@@ -40,7 +40,10 @@
   "devDependencies": {
     "chai": "^3.5.0",
     "chai-as-promised": "^5.2.0",
+    "fs-extra": "^0.26.7",
```
rwjblue: This isn't used in this PR

tomdale: It's used here: https://github.com/ember-fastboot/ember-fastboot-server/pull/39/files#diff-4b704b373a591b75fa93036c2f681941R95

rwjblue: @tomdale - That is using `fs-promise`, not `fs-extra`.

tomdale: @rwjblue You appear to be correct.

PR #41, opened 2016-04-13
Defer rendering via promise chaining

danmcclain:
There are no tests on this side, but there are in https://github.com/ember-fastboot/ember-cli-fastboot/issues/169

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -86,6 +86,13 @@ EmberApp.prototype.visit = function(path, options) {
     .then(function(instance) {
       return instance.visit(path, bootOptions);
     })
+    .then(function(instance) {
+      var fastbootInfo = instance.lookup('info:-fastboot');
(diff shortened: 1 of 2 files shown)


lib/ember-app.js
```diff
@@ -86,6 +86,13 @@ EmberApp.prototype.visit = function(path, options) {
     .then(function(instance) {
       return instance.visit(path, bootOptions);
     })
+    .then(function(instance) {
+      var fastbootInfo = instance.lookup('info:-fastboot');
```
tomdale: Don't we already have the FastBootInfo object, since we provide it to the app? Why do we have to look it up again?

danmcclain: We don't have the FastBootInfo object in this context, it was created and registered within the instance in `registerFastBootInfo`, then "forgotten"

PR #45, opened 2016-04-21
Implement the remaining Headers spec in FastBootHeaders

danmcclain:
https://developer.mozilla.org/en-US/docs/Web/API/Headers

### DIFF
--- a/lib/fastboot-headers.js
+++ b/lib/fastboot-headers.js
@@ -1,7 +1,38 @@
 // Partially implements Headers from the Fetch API
 // https://developer.mozilla.org/en-US/docs/Web/API/Headers
 function FastBootHeaders(headers) {
-  this.headers = headers
+  headers = headers || {};
+  this.headers = {};
+
+  for (var header in headers) {
+    this.headers[header] = headers[header].split(', ');
+  };
+}
+
+FastBootHeaders.prototype.append = function(header, value) {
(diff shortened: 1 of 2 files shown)


lib/fastboot-headers.js
```diff
@@ -1,25 +1,79 @@
...
 // https://developer.mozilla.org/en-US/docs/Web/API/Headers
 function FastBootHeaders(headers) {
-  this.headers = headers
+  headers = headers || {};
+  this.headers = {};
+
+  var _this = this;
+  Object.keys(headers).forEach(function(header) {
+    _this.headers[header] = headers[header].split(', ');
+  });
```
rwjblue: you can pass `this` as a second arg to avoid the `_this` variable above

lib/fastboot-headers.js
```diff
@@ -1,25 +1,79 @@
...
+  var entries = [];
+
+  for(var key in this.headers) {
+    this.headers[key].forEach(function(value) {
+      entries.push(key);
+    });
   }
 
-  return [];
+  return entries[Symbol.iterator]();
```
rwjblue: Can we use `Symbol.iterator` here?  What is our lowest Node version supported?

danmcclain: The tests passed on all versions, so I'd say yes? 💃 @rwjblue

PR #46, opened 2016-04-22
FastBootWatcher (ember-fastboot --watch)

pwfisher:
Adds `--watch` option which triggers reload on changes within `distPath`
Fixes cli-test leaving server alive when there is an exception
Only spawns one watcher (`if cluster.isMaster`)
Debounces watcher events

Inspired by https://github.com/ember-cli/ember-cli/blob/64f5a90/lib/models/server-watcher.js

### DIFF
--- a/bin/ember-fastboot
+++ b/bin/ember-fastboot
@@ -10,6 +10,9 @@ var expressCluster = require('express-cluster');
 var cluster = require('cluster');
 var parseArgs = require('minimist');
 
+var FastBootWatcher = require('../lib/watcher');
+var debounce = require('lodash.debounce');
+
 var argOptions = {
   default: { port: 3000, host: '::' }
 };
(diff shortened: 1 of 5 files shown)


rwjblue:
This looks like really great work, thank you!  

I am a bit concerned that this seems to be the wrong "layer" for watching. ember-fastboot-server already supports reloading when it receives a `SIGUSR1` signal and I think sending that signal from a deploy tool (or even from within ember-cli-fastboot) is the right way to go for this.

pwfisher:
You're right, it should be simple to add `SIGUSR1` to the `ember-cli-fastboot` build, but AFAIK it won't solve anything for us. We do not use `ember fastboot` to launch our fastboot server, even in dev. (Our fastboot server runs in a docker container inside a vagrant box and mounts `dist` from the host. Our `ember build --watch` runs on the host.)

Perhaps this should become `fastboot-watch-notifier` for `fastboot-app-server`?

The benefit of this PR is that it adds watch support:
 1. no matter how you use `ember-fastboot-server`
 2. in a way that is simple to understand and easy to use.

On the downside, it adds a dependency (`sane`) which is not really needed in production.

PR #49, opened 2016-04-27
Update Node version matrix in Travis config.

rwjblue:
Node 0.11 should not be supported, and we should test against current stable version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: node_js
 node_js:
-  - "4.1"
-  - "4.0"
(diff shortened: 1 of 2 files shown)


danmcclain:
Somehow breaks the build....

rwjblue:
Tracked down the issue with Node 6.0.0, and fixed in https://github.com/ember-fastboot/ember-fastboot-server/pull/49/commits/66e827a7ecc7b55cf6ec4949ef4c6f0a4e59ff80.

PR #50, opened 2016-04-27
Fix FastBoot require

danmcclain:
FastBoot.require assumed that the package was a npm module, this
attempts to load the built in module before checking node_modules

### DIFF
--- a/lib/ember-app.js
+++ b/lib/ember-app.js
@@ -3,6 +3,7 @@ var path = require('path');
 
 var SimpleDOM = require('simple-dom');
 var najax = require('najax');
+var existsSync = require('exists-sync');
 var debug   = require('debug')('ember-cli-fastboot:ember-app');
 var FastBootInfo = require('./fastboot-info');
 var VMSandbox = require('./sandboxes/vm');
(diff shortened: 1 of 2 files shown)


lib/ember-app.js
```diff
@@ -153,7 +153,13 @@ function serializeHTML(doc, rootElement) {
 function buildWhitelistedRequire(whitelist, distPath) {
   return function(moduleName) {
     if (whitelist.indexOf(moduleName) > -1) {
-      return require(path.join(distPath, 'node_modules', moduleName));
+      try {
+        // Attempt to see if it's a built-in module
+        return require(moduleName);
+      } catch (e) {
+        // See if it's an npm module, since it wasn't built in
+        return require(path.join(distPath, 'node_modules', moduleName));
```
rwjblue: I'd rather resolve from this location instead (via resolve package). Also, is distPath guaranteed to have a leading '.' or '/'? If not then I'm not sure his will work..

danmcclain: `path.resolve` is called here: https://github.com/ember-fastboot/ember-fastboot-server/blob/master/lib/server.js#L53

rwjblue: Chatted with @danmcclain in Slack, I think he is going to work on removing the `try` / `catch` if we can...

danmcclain:
@rwjblue try/catch removed

PR #52, opened 2016-04-29
Add ENV to fastboot-info

TeknoFiend:
We'd like to be able to pass ENV information to the Ember app so we can use the same build in production & staging (pre-production) but with different API keys/hostnames/etc.

This is a pretty brain-dead method of doing that but it works for us. Hopefully it's at least a starting point for a discussion.

### DIFF
--- a/lib/fastboot-info.js
+++ b/lib/fastboot-info.js
@@ -11,6 +11,7 @@ function FastBootInfo(request, response, hostWhitelist) {
   this.deferredPromise = RSVP.resolve();
   this.request = new FastBootRequest(request, hostWhitelist);
   this.response = new FastBootResponse(response);
+  this.ENV = process.env;
 }
 
 FastBootInfo.prototype.deferRendering = function(promise) {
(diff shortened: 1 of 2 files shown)


tomdale:
I think the work that @danmcclain did to make `config`… well… configurable should handle this use case. I am a little leery of passing the entire ENV into the running Ember app.

@danmcclain We should think about if there's some way to easily use environment variables as values in the configuration. Having to specify the entire config as a JSON string kinda sucks if you want to change a single configuration setting.

TeknoFiend:
Thanks for commenting. Does anyone have a pointer to said work? 

We want to configure _without_ rebuilding because so we can test what actually goes into production.  I don't believe `storeConfigInMeta` supports something like that.

The ENV is a natural place for this, though I agree that passing the whole object in is not ideal. A whitelist  or a nested object (like `emberServerENV`) would work fine.

tomdale:
@TeknoFiend See https://github.com/ember-fastboot/fastboot/pull/58 for the work @danmcclain did to let you set the app's config as an env var. The only downside of this work is that you have to set the _whole_ config at once, rather than tweaking just a few properties.

PR #56, opened 2016-05-05
Expose config as a function on the FastBoot global

danmcclain:
Allow for config changes at deploy time via the `package.json` file. Allow for config to be set as an environment variable


lib/server.js
```diff
@@ -49,12 +49,14 @@ FastBootServer.prototype.buildEmberApp = function(distPath, ui) {
     this.app.destroy();
   }
 
+
```
rwjblue: ✂️

rwjblue:
This looks nice.  This is the precursor to a PR in ember-cli-fastboot that will allow us to remove `storeConfigInMeta` stuff.  Once all of that lands, it will be possible to "promote" the same assets from one env to another without recompiling.

test/fastboot-server-test.js
```diff
@@ -1,9 +1,30 @@
...
+    server.stop();
+  }
+}
+
+function requestIndex(assertion) {
+  return function(info) {
+    var url = 'http://[' + info.host + ']:' + info.port + '/';
+
+    return request(url)
+    .then(function(html) {
```
rwjblue: this can become:

```js
return request(url)
  .then(assertion)
  .then(function() {
    return info;
  });
```

test/fastboot-server-test.js
```diff
@@ -35,51 +56,164 @@ describe("FastBootServer", function() {
...
+    var promise = server.start()
+      .then(requestIndexRoute)
+      .finally(cleanup(server));
+    delete process.env.APP_CONFIG;
+
+    return promise;
 
-    function requestFirstApp(info) {
-      url = 'http://[' + info.host + ']:' + info.port + '/';
+    function requestIndexRoute(info) {
```
rwjblue: Can this be removed in favor of the `requestIndex`  helper function added above?

PR #57, opened 2016-05-05
Major refactor, rename to fastboot

tomdale:
**Not ready for merge**

This PR tracks the work to make some improvements to the library formerly known as `ember-fastboot-server`, as well significantly decrease its scope.

The major changes in this PR are:

1. The results of calling `visit()` are now fully encapsulated in a `Result` type, making the developer ergonomics much better.
2. The CLI (`$ ember-fastboot dist/path/`) and middleware functionality have been removed.

Now, it's easy to write your own code to render an Ember app on the server:

```js
const FastBoot = require('fastboot');

let fastboot = new FastBoot({
  distPath: 'path/to/dist'
});

fastboot.visit('/photos')
  .then(r => r.html())
  .then(html => res.send(html));
```

Additionally, this PR adds a `resilient` flag to the FastBoot class, which tells it not to treat application errors as promise rejections. This makes it easy for a consuming HTTP server to send the normal, "empty" HTML page back to the client, so they can try to recover from the error client-side, rather than getting a 500 error.

```js
let fastboot = new FastBoot({
  distPath: 'path/to/dist',
  resilient: true
});
```

The error is still available on the `Result`:

```js
fastboot.visit('/photos')
  .then(r => {
    if (r.error) {
      console.log('Error rendering: ', r.error.stack);
    }

    return r.html().then(html => res.send(html));
  });
```

Next steps are:

* [ ] Get this PR into a state ready to merge
* [ ] Migrate to `fastboot` name
  * [ ] Move repo on GitHub
  * [ ] Publish updated package on npm
  * [ ] Add deprecation warning to current branch
  * [ ] Migrate CLI command to `fastboot-app-server`
  * [ ] Migrate middleware functionality to new `fastboot-express-middleware` package


README.md
```diff
@@ -52,35 +62,7 @@ rendering your Ember.js application using the `ember-fastboot` command:
 $ ember-fastboot path/to/dist --port 80
```
danmcclain: The command line instructions should be removed, right?

PR #58, opened 2016-05-08
Expose config as a function on the FastBoot global

danmcclain:
Allow for config changes at deploy time via the `package.json` file
Allow for config to be set as an environment variable

This is the same as #56, but specifically against `fastboot-rename`


rwjblue:
Hmm, didn't we land this already?

danmcclain:
@rwjblue this one is specifically against the rename branch

rwjblue:
Ahh, I see.  Thanks for the clarification!

PR #59, opened 2016-05-09
Fast Boot "Shoebox": A means to serialize application state into DOM for consumption by the browser

habdelra:
Serialize the application state into DOM so that you can pass application state from the fastboot rendered app to the browser rendered app

To do:
- [x] use custom script tag `type` attribute to prevent browser from parsing as javascript
- [x] write out a script DOM node for each key that is serialized, and namespace the serialization key
- [x] don't use "serialize" or "deserialize" as the names for things, come up with a new cute name for this.


habdelra:
@tomdale ok I'm ready for your 👀 now

danmcclain:
Tests please! I also added comments about using id instead of data attributes on the cli side, as browsers have optimized id retrieval

habdelra:
Will do. Thanks

habdelra:
@danmcclain I've updated with your suggestions. thanks!

PR #70, opened 2016-06-16
Exclude tests from shipped npm package

pwfisher:
```
$ du -sh node_modules/fastboot*/test
5.7M	node_modules/fastboot-app-server/test
 17M	node_modules/fastboot-express-middleware/test
 27M	node_modules/fastboot/test
```


rwjblue:
Good catch!

PR #71, opened 2016-07-08
Only access instance.getURL if the instance has booted

jasonmit:
Accessing it if the instance failed to boot causes [Ember's router to try and access `location.getURL`](https://github.com/ember-fastboot/fastboot/pull/71) when `location` is a string (`"none"`) instead the expected `Ember.NoneLocation` instance.

This is leading to all uncaught errors that occur during the app boot to be masked/swallowed.

https://github.com/ember-fastboot/ember-cli-fastboot/issues/223

I can send a fix into Ember to not invoke`getURL` when the location object was not created.

### DIFF
--- a/lib/result.js
+++ b/lib/result.js
@@ -10,6 +10,7 @@ const HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
  */
 class Result {
   constructor(options) {
+    this.instanceBooted = false;
     this._doc = options.doc;
     this._html = options.html;
     this._fastbootInfo = options.fastbootInfo;
@@ -71,7 +72,9 @@ class Result {
   }
 
   _finalizeMetadata(instance) {
-    this.url = instance.getURL();
+    if (this.instanceBooted) {
+      this.url = instance.getURL();
+    }
 
     let response = this._fastbootInfo.response;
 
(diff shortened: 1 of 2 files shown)


lib/result.js
```diff
@@ -70,8 +70,10 @@ class Result {
     return this;
   }
 
-  _finalizeMetadata(instance) {
-    this.url = instance.getURL();
+  _finalizeMetadata(instance) {
+    if (this.booted) {
```
jasonmit: I went with this approach versus using `instance._booted` since it's marked private

rwjblue:
This seems good to me, but I agree that we should probably also make some changes upstream in Ember around this too.

thoov:
There are some other instances where the getURL error is still happening.  I plan to work through those cases this weekend (@jasonmit on @thoov's account)

jasonmit:
@rwjblue okay, should be fine and ready to merge when you're ready.  I'll also start on fixing the upstream bug in Ember.

et:
@jasonmit - did you ever get a chance to work on the upstream bug in Ember? I'm starting to see this in an acceptance test. Let me know if I can help too 😄

seanmcc:
I just got this error. In my case, the problem was not in an initializer, but a bad route definition. The way I figured this out was to start ember server without fastboot `FASTBOOT_DISABLED=true ember serve`. The error I saw in the browser was very useful, but for some reason, fastboot errors can be opaque.

PR #74, opened 2016-07-19
0.12 support

tomdale:
This PR adds back support for Node 0.12. It works by shipping a transpiled version of the source code to npm, and when a user installs it in the Node 0.12 environment, switches to that version.

The major constraints we had were:

1. We wanted to use new syntax features in ES2015, particularly arrow functions, which really help clean up heavily async code.
2. We still wanted to support Node 0.12.
3. We didn't want to slow down development by requiring a transpilation step after every code change, like to run the tests, for example.

This PR makes the following changes:

1. Source is moved into the `src/` directory.
2. Source should only use ES2015 syntax available in Node 4 or later, although if newer syntax is used, it will still fall back to the transpiled output.
3. The [broccoli-module-alchemist](https://github.com/monegraph/broccoli-module-alchemist) plugin is used to transpile source into CommonJS + ES5 and placed into the `dist/` directory.
4. [broccoli-module-alchemist-install](https://github.com/tomdale/broccoli-module-alchemist-install) is used to switch to the right version at package install time.
5. To run the tests in older versions of Node, Babel integration is added to Mocha. This was mostly for convenience, and to avoid the transpilation step during test runs.
6. Inside the tests, the `broccoli-module-alchemist` version of `require()` is used to require modules from the package; this will grab the file from either `src/` or `dist/cjs/` as appropriate at runtime, allowing the tests to pass on Node 0.12.


danmcclain:
LGTM 👍

tomdale:
@nathanhammond Sounds like the best Christmas present ever. ;)

PR #75, opened 2016-08-01
Add metadata option for .visit

thoov:
Allows an option of metadata to be passed into options like:
```js
fastboot.visit('...', { metadata: { foo: 'bar' } });
```

This enables per request information to be passed into the app such as user session info.

Related service change: https://github.com/ember-fastboot/ember-cli-fastboot/pull/245

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -202,7 +202,7 @@ class EmberApp {
     let fastbootInfo = new FastBootInfo(
       req,
       res,
-      { hostWhitelist: this.hostWhitelist, metaData: options.metaData }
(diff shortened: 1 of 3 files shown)


jasonmit:
@danmcclain @tomdale this is based on what we discussed last week on Slack.

We have an application that builds up state prior to the fastboot-middleware.  So rather than rebuilding that state up within the fastboot app, we would like the ability to provide that to the application instance.

We currently do this and think it may be beneficial for future travelers as well to have the option.

kellyselden:
I just became blocked on this.

PR #79, opened 2016-08-17
Shoebox encoding

pwfisher:
SimpleDOM [escapes `&`, `<` and `>`](https://github.com/ember-fastboot/simple-dom/blob/80ecb54/lib/simple-dom/html-serializer.js#L46-L57) when it serializes text nodes, so our Shoebox payload gets garbled. Use [`createRawHTMLSection`](https://github.com/ember-fastboot/simple-dom/blob/80ecb54/lib/simple-dom/html-serializer.js#L63-L65), instead.

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -311,7 +311,7 @@ function createShoebox(doc, fastbootInfo) {
 
     let value = shoebox[key];
     let textValue = JSON.stringify(value);
-    textValue = textValue.replace(/<\/(script)/i, '<\\/$1');
+    textValue = textValue.replace(/<\/(script[ />])/i, '<\\/$1');
 
     let scriptText = doc.createRawHTMLSection(textValue);
     let scriptEl = doc.createElement('script');
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -310,7 +310,10 @@ function createShoebox(doc, fastbootInfo) {
     if (!shoebox.hasOwnProperty(key)) { continue; }
 
     let value = shoebox[key];
-    let scriptText = doc.createRawHTMLSection(JSON.stringify(value));
+    let textValue = JSON.stringify(value);
+    textValue = textValue.replace('</script', '<\\/script');
+
+    let scriptText = doc.createRawHTMLSection(textValue);
     let scriptEl = doc.createElement('script');
 
     scriptEl.setAttribute('type', 'fastboot/shoebox');
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -310,7 +310,7 @@ function createShoebox(doc, fastbootInfo) {
     if (!shoebox.hasOwnProperty(key)) { continue; }
 
     let value = shoebox[key];
-    let scriptText = doc.createTextNode(JSON.stringify(value));
+    let scriptText = doc.createRawHTMLSection(JSON.stringify(value));
     let scriptEl = doc.createElement('script');
 
     scriptEl.setAttribute('type', 'fastboot/shoebox');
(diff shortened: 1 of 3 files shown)


rwjblue:
Hmm. Does this point to a larger issue with escaping?

I.e. what would the following do after your changes:

```js
let shoebox = this.get('fastboot.shoebox');
shoebox.put('my-store', {
  foo: '</script>'
});
```

pwfisher:
Ouch. Excellent point.
- As a security issue, it would be a big deal. If users serialize models with user-provided input, we've just created a vulnerability.
- As a gotcha, it also creates a pretty nasty failure mode if you happen to serialize data with that string.

Good news, though, is that `</script` is *the* special case we need to worry about, so it's not a larger issue.

We should be fine with
```
let textValue = JSON.stringify(value);
textValue = textValue.replace('</script', '<\\/script');
let scriptText = doc.createRawHTMLSection(textValue);
```

Note the lack of a trailing `>`, as `</script foo>` is valid.

Look good to you?

cc @briangonzalez @arjansingh

src/ember-app.js
```diff
@@ -310,7 +310,10 @@ function createShoebox(doc, fastbootInfo) {
     if (!shoebox.hasOwnProperty(key)) { continue; }
 
     let value = shoebox[key];
-    let scriptText = doc.createTextNode(JSON.stringify(value));
+    let textValue = JSON.stringify(value);
+    textValue = textValue.replace('</script', '<\\/script');
```
jasonmit: This will not cover `</ScRiPt`

pwfisher: Wow. Duh. Thank you for catching that.

pwfisher:
So, is this really enough?

Let's get a real answer from the [HTML5 script element specs](https://www.w3.org/TR/html-markup/script.html):
- permitted contents: [Non-replaceable character data](https://www.w3.org/TR/html-markup/syntax.html#non-replaceable-character-data)
  - Non-replaceable character data can contain the following:
    - text, optionally including `<` characters
    - ambiguous ampersands
  - Non-replaceable character data has the following restrictions:
    - **must not contain character references**
    - **must not contain any occurrences of the string `</`, followed by characters that are a case-insensitive match for the tag name of the element containing the replaceable character data (for example, `</script` or `</style`, followed by a space character, `>`, or `/`.**
- Additional constraint:
  - A script element must not contain an occurrence of the character sequence `<!--` without a matching later occurrence of the character sequence `-->`.

**TL;DR**: Don't encode script tag content. Case-insensitive `</script` is a specifically prohibited special case.

Our matching is safely stricter than the specified `/<\/script[ />]/` (followed by space, `/` or `>`).

Re @rwjblue (in slack) "I think in general I'd prefer to do some sort of encoding/decoding"
- The spec says that script tag content "must not contain [character references](https://www.w3.org/TR/html-markup/syntax.html#syntax-charref)". That includes named (`&gt;`), decimal (`&#8224;`) and hex (`&#x2020`) character references.
- We could url-encode, but it is not required. It would be unexpected.

The simple solution works. Let's keep it simple.

chancancode:
If I understand this correctly, you should probably just do this: http://api.rubyonrails.org/classes/ERB/Util.html#method-c-json_escape

src/ember-app.js
```diff
@@ -310,7 +310,10 @@ function createShoebox(doc, fastbootInfo) {
     if (!shoebox.hasOwnProperty(key)) { continue; }
 
     let value = shoebox[key];
-    let scriptText = doc.createTextNode(JSON.stringify(value));
+    let textValue = JSON.stringify(value);
+    textValue = textValue.replace(/<\/(script[ />])/i, '<\\/$1');
```
tomdale: Doesn't this need a `g` flag? I think this will only catch the first instance of `</script`.

tomdale:
Closed in favor of #85.

PR #84, opened 2016-09-27
handle headers as both strings and arrays

kellyselden:
When I don't supply the express response in the visit options, everything works fine, but when I do supply it:

``` js
app.visit(url, {
  response: res
})
```

The headers parsing fails because sometimes the header values are already arrays and don't have the split function. A sample while debugging:

```
// values masked...
debug> exec headers
{ 'set-cookie': 
   [ 'ident=14xxxxxxxxx24; Max-Age=1800; Doma... (length: 160)',
     'session=0xxxxxxxxD; Domain=test.com; Path=/;... (length: 89)' ],
  'x-session-id': '5xxxxxxxxx4',
  'x-registration-id': '8xxxxxxx4',
  'x-appserver-id': 'test',
  token: 'eyJhbGciOiJIUzxxxxxxxxxxxxxxxxxRDVGQkR... (length: 1636)' }
```

I don't yet know if this is an issue only I would have for some reason, or it will start turning up in other apps.

### DIFF
--- a/src/fastboot-headers.js
+++ b/src/fastboot-headers.js
@@ -5,7 +5,14 @@ function FastBootHeaders(headers) {
   this.headers = {};
 
   for (var header in headers) {
-    this.headers[header] = headers[header].split(', ');
+    let value = headers[header];
+
+    // convert to array if not already
+    if (typeof value === 'string') {
+      value = value.split(', ');
+    }
+
+    this.headers[header] = value;
   }
 }
 
--- a/test/fastboot-headers-test.js
+++ b/test/fastboot-headers-test.js
@@ -68,6 +68,15 @@ describe('FastBootHeaders', function() {
     expect(headers.has('host')).to.be.false;
   });
 
+  it('handles header values that are already an array', function() {
+    var headers = {
+      'x-test-header': ['value1', 'value2']
+    };
+    headers = new FastBootHeaders(headers);
+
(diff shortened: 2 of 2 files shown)


kellyselden:
I should add that I'm using the standalone fastboot package, without any middleware.

src/fastboot-headers.js
```diff
@@ -5,7 +5,14 @@ function FastBootHeaders(headers) {
   this.headers = {};
 
   for (var header in headers) {
-    this.headers[header] = headers[header].split(', ');
+    let value = headers[header];
+
+    // convert to array if not already
+    if (value.split) {
```
arjansingh: I'd prefer doing an `instanceof` or `typeof` check here instead.

kellyselden: changed.

kellyselden:
I finally found the evidence needed to get this merged. Express turns the `set-cookie` response header into an array if you set more than one cookie:

https://github.com/expressjs/express/blob/bd47aeb88d28c6cff6e2d8e5378ec73c262e039f/lib/response.js#L805
https://github.com/expressjs/express/blob/bd47aeb88d28c6cff6e2d8e5378ec73c262e039f/lib/response.js#L682

kellyselden:
cc @danmcclain and @tomdale, I remember on the call you two expressed concern over this change.

rwjblue:
closed by https://github.com/ember-fastboot/fastboot/pull/113

PR #85, opened 2016-09-30
JSON escape rendered shoebox content

tomdale:
Once rendering is completed, the contents of the shoebox is converted into a string of JSON and inserted into the HTML output that is sent back to the browser.

However, because JSON and HTML content are mixed, there is the potential for security vulnerabilities. Specifically, if an attacker can cause an application to place user-generated content into the shoebox, that content could trick the browser into thinking script parsing had ended, and evaluate arbitrary code in the origin of the host.

For example, if an untrusted user could supply an article with the title of `</script><script>alert("owned")</script>`, the naive interpolation of that into the shoebox might look like:

```html
<script type="fastboot/shoebox" id="shoebox-article">
{"article":{"title":"</script><script>alert("owned")</script>"}}
</script>
```

In this case, the browser would interpret the `</script>` inside the JSON string as a real closing `script` tag, and thus would allow the attacker's code to execute in the application's origin ("XSS").

Upon examining the HTML5 parser specification, [we can observe that there is one, and only one, way to exit the "script data" state][spec]: the existence of a `<` character, which moves the state machine into the "script data less-than sign state". From the "script data less-than sign state", there are several more states that can be traversed through, and it requires the creation of a temporary buffer.

[spec]: https://www.w3.org/TR/html5/syntax.html#script-data-state

Thus we can conclude that the simplest, most effective way to prevent inadvertent end-of-script situations is to prevent the `<` character from ever appearing in shoebox content. If you never leave the "script data" state, you can feel fairly certain that you have prevented this particular vector of XSS attacks.

The good news is that this is easily accomplished. Both the JavaScript specification and the JSON specification allow for [Unicode escape sequences](https://mathiasbynens.be/notes/javascript-escapes#unicode).

Before insertion into the HTML document, we can replace characters that could be ambiguous to the HTML parser and replace them with Unicode escape sequences. These are no different from the unescaped values to the eyes of the JSON or JavaScript parser, but give us a high degree of confidence that the HTML parser will not attempt to treat them as anything other than script data.

This commit Unicode escapes the following characters:

* `<` and `>`, to prevent ambiguity with opening and closing tags.
* `&`, to prevent ambiguity with HTML entities.
* `\u2028` and `\u2029`, Unicode line/paragraph separators, which the JSON parser and JavaScript parser treat differently and thus can lead to mismatched data if JavaScript is used as the JSON parser.

This PR is based on @pwfisher's #79 but uses an updated approach that the core team feels is more robust to potential attack vectors.

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -310,7 +310,10 @@ function createShoebox(doc, fastbootInfo) {
     if (!shoebox.hasOwnProperty(key)) { continue; }
 
     let value = shoebox[key];
-    let scriptText = doc.createTextNode(JSON.stringify(value));
+    let textValue = JSON.stringify(value);
+    textValue = escapeJSONString(textValue);
+
+    let scriptText = doc.createRawHTMLSection(textValue);
     let scriptEl = doc.createElement('script');
 
     scriptEl.setAttribute('type', 'fastboot/shoebox');
@@ -320,6 +323,22 @@ function createShoebox(doc, fastbootInfo) {
   }
 
   return RSVP.resolve();
+}
+
+const JSON_ESCAPE = {
+  '&': '\\u0026',
+  '>': '\\u003e',
+  '<': '\\u003c',
+  '\u2028': '\\u2028',
+  '\u2029': '\\u2029'
+};
+
+const JSON_ESCAPE_REGEXP = /[\u2028\u2029&><]/g;
+
+function escapeJSONString(string) {
+  return string.replace(JSON_ESCAPE_REGEXP, function(match) {
+    return JSON_ESCAPE[match];
+  });
 }
 
 /*
--- a/test/fastboot-shoebox-test.js
+++ b/test/fastboot-shoebox-test.js
@@ -10,7 +10,7 @@ const FastBoot       = alchemistRequire('index');
 
 describe("FastBootShoebox", function() {
 
-  it("can render the shoebox HTML", function() {
+  it("can render the escaped shoebox HTML", function() {
     var fastboot = new FastBoot({
       distPath: fixture('shoebox')
     });
@@ -20,6 +20,11 @@ describe("FastBootShoebox", function() {
       .then(html => {
         expect(html).to.match(/<script type="fastboot\/shoebox" id="shoebox-key1">{"foo":"bar"}<\/script>/);
         expect(html).to.match(/<script type="fastboot\/shoebox" id="shoebox-key2">{"zip":"zap"}<\/script>/);
+
+        // Special characters are JSON encoded, most notably the </script sequence.
+        expect(html).to.include('<script type="fastboot/shoebox" id="shoebox-key4">{"nastyScriptCase":"\\u003cscript\\u003ealert(\'owned\');\\u003c/script\\u003e\\u003c/script\\u003e\\u003c/script\\u003e"}</script>');
+
+        expect(html).to.include('<script type="fastboot/shoebox" id="shoebox-key5">{"otherUnicodeChars":"\\u0026\\u0026\\u003e\\u003e\\u003c\\u003c\\u2028\\u2028\\u2029\\u2029"}</script>');
       });
   });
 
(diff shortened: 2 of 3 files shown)


habdelra:
This looks very very terrific. I have a son, he's 10 years old. He has computers, he's so good with these computers it's unbelievable. The security aspect of cyber is very very tough. And maybe it's hardly doable. But I think you have solved it @tomdale.

chancancode:
@tomdale FWIW, I _think_ escaping `\u2028` and `\u2029` only matters if you eval the content (as opposed to passing it into `JSON.parse`), but it probably have the pleasant side-effect of making things less confusing looking in the HTML.

Should we be unit-testing this, like actually passing the string into JSON.parse and assert that it comes out the same? We can probably just port these tests from Rails over: https://github.com/rails/rails/blob/b326e82dc012d81e9698cb1f402502af1788c1e9/actionview/test/template/erb_util_test.rb#L46-L62

mathiasbynens:
@chancancode 

> FWIW, I think escaping `\u2028` and `\u2029` only matters if you `eval` the content (as opposed to passing it into `JSON.parse`)

This is correct. It doesn’t make sense to include U+2028 and U+2029 in this patch.

> , but it probably have the pleasant side-effect of making things less confusing looking in the HTML.

The same can be said for any symbol that is not a printable ASCII symbol. I use [jsesc](https://github.com/mathiasbynens/jsesc) anywhere I generate JSON output at build time for this reason.

In my experience, U+2028 and U+2029 are very rarely used though — other weird non-ASCII characters are much more common.

mathiasbynens:
FWIW, I’ve looked into this problem before, and wrote down my findings + the escaping requirements for each solution: [_“Hiding JSON-formatted data in the DOM with CSP enabled”_](https://mathiasbynens.be/notes/json-dom-csp).

danmcclain:
From https://github.com/ember-fastboot/ember-cli-fastboot/issues/275

>The Result of a Fastboot request gets serialized by the SimpleDOM.HTMLSerializer before it is sent down to the client (https://github.com/ember-fastboot/fastboot/blob/master/src/result.js). This happens after the Shoebox is appended to the document, and has the unintended consequence of encoding any characters in the shoebox data that happen to be HTML entities. This includes characters such as &, <, >, ", `, and ', which are encoded into &amp;, &lt;, &gt;, &quot;, &#96; and &#x27;, respectively.

>This results in a data payload fetched on the server and stored in the Shoebox to be different than the same data payload fetched on the client. This can cause issues when rendering any data provided by the Shoebox since Ember's rendering layer treats these encoded HTML entities as strings rather than HTML. For example my API may return Broadway & 1st Ave, but if that data is fetched on the server and passed to the client in the Shoebox it will be provided as Broadway &amp; 1st Ave.

>Seems like the shoebox#retrieve method ought to have a step that decodes any of these entities before the data is passed to the application.



>It seems like there should be an idiomatic way to unescape these characters once they are fetched from the shoebox in the client. Characters &, <, and > will commonly occur in text data stored in the shoebox. When normally consumed from a REST API or other data source they won't be escaped and will display normally in a template, but when consumed from the shoebox they will be escaped and won't display correctly. This seems like a source of frustration and bugs. It seems like the escaping/unescaping of the shoebox content should be transparent to the developer and happen behind the scenes within the shoebox#retrieve method.

>This, however, may conflict with shoebox content where the original data had escaped entities that shouldn't actually be unescaped. Considering this, maybe it makes more sense to leave it up to the developer to unescape shoebox content as needed. If so, this should be mentioned in the docs, and maybe there ought to be a standard/idiomatic way to do the unescaping such as an alternate method to shoebox#retrieve that unescapes the content.

rwjblue:
I replied to the last comment in detail in https://github.com/ember-fastboot/ember-cli-fastboot/issues/275.

Also, created [this demo](http://jsbin.com/jojalo/edit?html,js,output) to show the result of this escaping before writing to the DOM does not affect the resulting parsed values when it is read (as mentioned in this PR's description).

rwjblue:
@tomdale - What is blocking this?  Seems like it closes a fairly important gap/issue, I'd like to land it ASAP if there are no objections...

tomdale:
@rwjblue This is high priority and agree it should have been merged already. There was some debate about whether or not we need to escape `\u2028` and `u2029`, but I don't think it does any harm to do. I've merged this PR as is but can remove those escapements in a future commit.

aexmachina:
I just downloaded the latest versions and I'm still getting incorrectly escaped JSON. I'm using `"ember-cli-fastboot": "^1.0.0-beta.12"` and `"fastboot-app-server": "^1.0.0-rc.5"`.

acorncom:
@aexmachina Looks like a new release hasn't been cut yet for this package or for `fastboot-app-server`

PR #91, opened 2016-10-18
Refactor Fastboot's visit API to take additional options

kratiahuja:
## Motivation
Fastboot today renders the application on the server side and exposes an API to serve the base page as well via [fastboot-app-server](https://github.com/ember-fastboot/fastboot-app-server). Various companies may chose to use their own proprietary infrastructure to serve the base page (with doing other things on top of it). While they serve the base page using a different infrastructure, they may want use this fastboot library with serving the  ember app on the server side as is.

This PR adds a bunch of options to the `FastBoot` so that we can use this library can be used as a standalone without needing to serve the base page to the browser using this library. It also now allows apps to be served in "routing-only" mode (aka bigpipe).

## Public API
In order to address the above, following are the additional options exposed via the `visit` method:
- [x] `metaData`: Allows to pass per request meta data (as `metaData`) for every `visit` and exposes it via the fastboot [service](https://github.com/ember-fastboot/ember-cli-fastboot/pull/280). This allows apps to pass in any additional data and use it in the app when routing/rendering on server side. 
- [x] `shouldRender`: Passes `shouldRender` option via `visit` which should allow the app to be executed in ["routing-only"](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-application/lib/system/application-instance.js#L378) mode. This can allow an app to disable rendering on per request basis and use bigpipe as we do at LinkedIn. It is more performant than a full render if you only plan to use the shoebox from server side rendered app.
- [x] `disableShoebox`: Allows apps to not send API data in the shoebox. Apps may be sending API data via different means.
- [x] `destroyAppInstanceInMs`: Allows apps to set an time in ms after which an app instance will forcefully be destroyed. This is a more for a failure mechanism and fixes this [issue](https://github.com/ember-fastboot/fastboot/issues/90)
- [x] `sandbox`: Creating a fastboot instance now allows you to pass a custom sandbox class. See [PR](https://github.com/ember-fastboot/fastboot/pull/92)
- [x] `addOrOverrideSandboxGlobals`: Allows to pass a object of extra variables and values that will be exposed in the sandbox. This could be very infra specific. See [PR](https://github.com/ember-fastboot/fastboot/pull/92)
- [x] Expose `domContents` method from the result object. This will allow an app using a different infrastructure to read this library and get the data of the rendered route. 

## TODO
- [ ] Update the fastboot-website with all of the above public API
- [ ] Update fastboot README.md with this API

## Notes
Once this PR is merged we need to close the following issues/ merge these PRs in order for an app to use any infrastructure to serve the base page and still use server side rendering perfomantly: 
1. We need the parallel PR to this to be merged. The PR is [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/280).
2. We need to fix the fastboot build performance issues. The related RFC in `ember-cli` is [here](https://github.com/ember-cli/ember-cli/issues/6350).
3. We need to expose an API during build time to allow addons to add other assets to be loaded in fastboot. This issue is being tracked [here](https://github.com/ember-fastboot/ember-cli-fastboot/issues/270).
4. We need the sandbox PR to also be merged. See [here](https://github.com/ember-fastboot/fastboot/pull/92)

Once the above issues are fixed in addition to this PR, at LinkedIn we will be able to simply use:
```javascript
  var fastbootServer = new FastbootServer({
     distPath: '/foo',
     sandbox: <mysandboxclass>,
     addOrOverrideSandboxGlobals: {...}
  });
  fastbootServer.visit(route, {
     shouldRender: false,
     metaData: {...},
     request: {
       headers: {...}
     },
     disableShoebox: false,
     destroyAppInstanceInMs: <some number>
  });
```
This will allow us to remove usage of all private APIs that will be initially used from this library.

cc: @tomdale @arjansingh @danmcclain @kristoferbaxter @ryanone @masonwan @melmerp @chadhietala

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "fs-promise": "^0.5.0",
     "mocha": "^2.4.5",
     "mocha-jshint": "^2.3.1",
-    "request-promise": "^2.0.1",
     "temp": "^0.8.3"
   }
 }
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -190,20 +190,29 @@ class EmberApp {
    * @param {string} path the URL path to render, like `/photos/1`
    * @param {Object} options
    * @param {string} [options.html] the HTML document to insert the rendered app into
+   * @param {Object} [options.metadata] Per request specific data used in the app.
+   * @param {Boolean} [options.shouldRender] whether the app should do rendering or not. If set to false, it puts the app in routing-only.
+   * @param {Boolean} [options.disableShoebox] whether we should send the API data in the shoebox. If set to false, it will not send the API data used for rendering the app on server side in the index.html.
+   * @param {Integer} [options.destroyAppInstanceInMs] whether to destroy the instance in the given number of ms. This is a failure mechanism to not wedge the Node process (See: https://github.com/ember-fastboot/fastboot/issues/90)
    * @param {ClientRequest}
+   * @param {ClientResponse}
    * @returns {Promise<Result>} result
    */
   visit(path, options) {
     let req = options.request;
     let res = options.response;
     let html = options.html || this.html;
+    let disableShoebox = options.disableShoebox || false;
+    let destroyAppInstanceInMs = options.destroyAppInstanceInMs;
 
-    let bootOptions = buildBootOptions();
+    let shouldRender = (options.shouldRender !== undefined) ? options.shouldRender : true;
+    let bootOptions = buildBootOptions(shouldRender);
     let fastbootInfo = new FastBootInfo(
       req,
       res,
       { hostWhitelist: this.hostWhitelist, metadata: options.metadata }
     );
+
     let doc = bootOptions.document;
 
     let instance;
@@ -213,6 +222,19 @@ class EmberApp {
       html: html,
       fastbootInfo: fastbootInfo
     });
+
+    let destroyAppInstanceTimer;
+    if (parseInt(destroyAppInstanceInMs, 10) > 0) {
+      // start a timer to destroy the appInstance forcefully in the given ms.
+      // This is a failure mechanism so that node process doesn't get wedged if the `visit` never completes.
+      destroyAppInstanceTimer = setTimeout(function() {
+        if (instance && !result.instanceDestroyed) {
+          result.instanceDestroyed = true;
+          result.error = new Error('App instance was forcefully destroyed in ' + destroyAppInstanceInMs + 'ms');
+          instance.destroy();
+        }
+      }, destroyAppInstanceInMs);
+    }
 
     return this.buildAppInstance()
       .then(appInstance => {
@@ -225,12 +247,22 @@ class EmberApp {
       .then(() => result.instanceBooted = true)
       .then(() => instance.visit(path, bootOptions))
       .then(() => waitForApp(instance))
-      .then(() => createShoebox(doc, fastbootInfo))
+      .then(() => {
+        if (!disableShoebox) {
+          // if shoebox is not disabled, then create the shoebox and send API data
+          createShoebox(doc, fastbootInfo);
+        }
+      })
       .catch(error => result.error = error)
       .then(() => result._finalize())
       .finally(() => {
-        if (instance) {
+        if (instance && !result.instanceDestroyed) {
+          result.instanceDestroyed = true;
           instance.destroy();
+
+          if (destroyAppInstanceTimer) {
+            clearTimeout(destroyAppInstanceTimer);
+          }
         }
       });
   }
@@ -274,14 +306,15 @@ class EmberApp {
  * Builds an object with the options required to boot an ApplicationInstance in
  * FastBoot mode.
  */
-function buildBootOptions() {
+function buildBootOptions(shouldRender) {
   let doc = new SimpleDOM.Document();
   let rootElement = doc.body;
 
   return {
     isBrowser: false,
     document: doc,
-    rootElement: rootElement
+    rootElement,
+    shouldRender
   };
 }
 
--- a/src/fastboot-info.js
+++ b/src/fastboot-info.js
@@ -6,11 +6,17 @@ var FastBootResponse = require('./fastboot-response');
  * A class that encapsulates information about the
  * current HTTP request from FastBoot. This is injected
  * on to the FastBoot service.
+ *
+ * @param {ClientRequest} the incoming request object
+ * @param {ClientResponse} the response object
+ * @param {Object} additional options passed to fastboot info
+ * @param {Array} [options.hostWhitelist] expected hosts in your application
+ * @param {Object} [options.metaData] per request meta data
  */
-function FastBootInfo(request, response, options = {}) {
-  const { hostWhitelist, metadata } = options;
+function FastBootInfo(request, response, options) {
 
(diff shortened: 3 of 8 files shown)


src/ember-app.js
```diff
@@ -210,6 +222,19 @@ class EmberApp {
       fastbootInfo: fastbootInfo
     });
 
+    let destroyAppInstanceTimer;
+    if (destroyAppInstanceInMs) {
+      // start a timer to destroy the appInstance forcefully in the given ms.
+      // This is a failure mechanism so that node process doesn't get wedged if the `visit` never completes.
+      destroyAppInstanceTimer = setTimeout(function() {
```
arjansingh: Personally I'm not a fan of using `setTimeout` server-side since it's not deterministic. I understand the rationale here but let's think about it to see if there's a better way of implementing this.

kratiahuja: On other way probably is to do a `Ember.run.debounce`  so that it is hooked on to Ember's run loop. I am open to using that at both places if needed but it requires some more refactoring to run this in the sandbox context.

src/ember-app.js
```diff
@@ -210,6 +222,19 @@ class EmberApp {
       fastbootInfo: fastbootInfo
     });
 
+    let destroyAppInstanceTimer;
+    if (destroyAppInstanceInMs) {
```
rwjblue: Just to confirm, a `undefined` / `null` / `0` values wait indefinitely, right?

src/ember-app.js
```diff
@@ -190,16 +190,28 @@ class EmberApp {
...
     let res = options.response;
+    let metaData = options.metaData;
     let html = options.html || this.html;
+    let disableShoebox = options.disableShoebox || false;
+    let destroyAppInstanceInMs = options.destroyAppInstanceInMs;
+    let hostWhitelist = this.hostWhitelist;
 
-    let bootOptions = buildBootOptions();
-    let fastbootInfo = new FastBootInfo(req, res, this.hostWhitelist);
+    let shouldRender = (options.shouldRender !== undefined) ? options.shouldRender : true;
```
rwjblue: Does `options.hasOwnProperty('shouldRender') ? options.shouldRender : true;` work better (I am unsure if it is better...)?

src/ember-app.js
```diff
@@ -270,14 +305,15 @@ class EmberApp {
...
-function buildBootOptions() {
+function buildBootOptions(shouldRender) {
   let doc = new SimpleDOM.Document();
   let rootElement = doc.body;
 
   return {
     isBrowser: false,
     document: doc,
-    rootElement: rootElement
+    rootElement: rootElement,
```
rwjblue: Can probably just use short hand here:

```js
  rootElement,
  shouldRender
```

src/fastboot-info.js
```diff
@@ -7,13 +7,15 @@ var FastBootResponse = require('./fastboot-response');
  * current HTTP request from FastBoot. This is injected
  * on to the FastBoot service.
  */
-function FastBootInfo(request, response, hostWhitelist) {
+function FastBootInfo(request, response, infoOptions) {
   this.deferredPromise = RSVP.resolve();
+  const { hostWhitelist, metaData } = infoOptions;
```
rwjblue: Feel free to disregard, but this is the only `const` here and I've generally seen us follow the "liberal-let" ([from here](http://madhatted.com/2016/1/25/let-it-be)), this might want to be a `let`?

src/fastboot-info.js
```diff
@@ -7,13 +7,15 @@ var FastBootResponse = require('./fastboot-response');
  * current HTTP request from FastBoot. This is injected
  * on to the FastBoot service.
  */
```
rwjblue: Should this document the constructor arguments?  Definitely non-blocking, but it might be nice to document what things are expected to be in `infoOptions`...

src/result.js
```diff
@@ -46,6 +47,24 @@ class Result {
   }
 
   /**
+   * Returns the serialized representation of the HEAD.
+   *
+   * @returns {String} serialized version of the DOM HEAD
+  */
+  getHead() {
```
rwjblue: Should this (and `getBody`) be a real getter instead? I ever so _slightly_ dislike the `get***` and `set***` pattern when we can just use native getters/setters if that works the same for us...

src/ember-app.js
kratiahuja: I find this `options.shouldRender !== undefined` much clear personally 😄

src/fastboot-info.js
kratiahuja: Yes it would be nice to add this.

src/result.js
kratiahuja: I prefer explicit API for getting private things from a class. getters/setters kind of make me think it can be misused :). But that is probably just me. Keeping this explicit API. 

Or I can provide a more verborse API like `domContents` which would  return an object like:
```javascript
{
   'head': head,
   'body': body
}
```

would you prefer something like that?

chadhietala:
Looks good to me.

src/ember-app.js
kratiahuja: Just to be safe I added a check of it being a number greater than 0 😄

src/ember-app.js

src/result.js
rwjblue: Yes, I like that better than my suggestion anyways :smiley_cat:

kratiahuja:
@rwjblue this is ready to be reviewed again if you need! Once you and everyone else give a go ahead, I'll squash my commits.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "fastboot",
   "version": "1.0.0-rc.0",
   "description": "Library for rendering Ember apps in node.js",
-  "main": "src/index.js",
+  "main": "dist/cjs/index.js",
```
rwjblue: I believe it is getting changed by https://github.com/monegraph/broccoli-module-alchemist, when it detects that you are running a version of node that isn't compatible with `src/` (I _think_ that Node 4.x and above would not have this issue).

kratiahuja: Aah ok good to learn this. I'll fix that in my next commit

rwjblue:
@tomdale - Any concerns here?

kellyselden:
I think #75 should be merged first, to get `metadata` instead of `metaData`, then this rebase on top. I believe that was the conclusion at the last meeting.

kratiahuja:
Squashed and rebased from remote master. This is ready to go once everyone gives a 👍

src/ember-app.js
```diff
@@ -190,20 +190,29 @@ class EmberApp {
    * @param {string} path the URL path to render, like `/photos/1`
    * @param {Object} options
    * @param {string} [options.html] the HTML document to insert the rendered app into
+   * @param {metaData} [options.metadata] Per request specific data used in the app.
```
kellyselden: should this be `@param {Object}`?

kratiahuja: Yeah my bad. will fix it.

rwjblue:
I'm happy with this, thank you @kratiahuja!

Will try to land and release tomorrow, unless there are objections.

PR #92, opened 2016-10-19
Allow custom sandbox to be used in fastboot.

kratiahuja:
Fastboot had an API to allow any app to use a custom sandbox class. However, this was never fully working. This PR fixes that.

This PR also adds a `addOrOverrideSandboxGlobals` which allows to pass any other variables that need to be exposed in the sandbox or allow existing variables to be overriden.

cc: @tomdale @rwjblue @danmcclain @ryanone @masonwan @arjansingh @kristoferbaxter


src/ember-app.js
```diff
@@ -54,23 +54,30 @@ class EmberApp {
    *
    * @param {string} distPath path to the built Ember app to load
    * @param {Sandbox} [sandboxClass=VMSandbox] sandbox class to use
+   * @param {Object} [sandboxGlobals={}] any additional variables to expose in the sandbox
    */
-  buildSandbox(distPath, sandboxClass) {
+  buildSandbox(distPath, sandboxClass, sandboxGlobals) {
```
rwjblue: Small suggestion here: can we name the argument `additionalSandboxGlobals`?  It makes it clearer to me that these augment the existing set of globals (not replace them).

src/ember-app.js
```diff
@@ -54,23 +54,30 @@ class EmberApp {
...
+    // add any additional user provided variables in the sandbox
+    let globals = {
+      najax: najax,
+      FastBoot: {
+        require: sandboxRequire,
+        config: appConfig
       }
+    };
+    for (let key in sandboxGlobals) {
+      globals[key] = sandboxGlobals[key];
```
rwjblue: We should clarify if this is intended to support clobbering built-in globals (i.e. a value of `{ najax: null }` with the current implementation would make `najax` not available in the sandbox).  If we do not intend to support clobbering we should check and throw if the same key is in both `additionalSandboxGlobals` and `globals` before setting...

src/index.js
```diff
@@ -36,15 +38,17 @@ class FastBoot {
    * @param {string} options.distPath the path to the built Ember application
    * @param {Boolean} [options.resilient=false] if true, errors during rendering won't reject the `visit()` promise but instead resolve to a {@link Result}
    * @param {Sandbox} [options.sandbox=VMSandbox] the sandbox to use
+   * @param {Object} [options.sandboxGlobals={}] any additional sandbox variables that an app server wants to expose
```
rwjblue: Same suggestion RE: naming here as above (`additionalSandboxGlobals`)

src/ember-app.js
kratiahuja: The idea was to override `{najax: null}` since we can't use `najax` in our system.

src/ember-app.js
```diff
@@ -54,23 +54,30 @@ class EmberApp {
...
+    }
```
kristoferbaxter: Would recommend considering an Object.assign here.

`const globals = Object.assign({
  najax,
  FastBoot: {
    require: sandboxRequire,
    config: appConfig
  }
}, sandboxGlobals);

return new Sandbox({
  globals
});`

src/index.js
kratiahuja: Fixed to `addOrOverrideSandboxGlobals`

src/ember-app.js
kratiahuja: @kristoferbaxter `Object.assign` isn't present in Node 0.12. I wouldn't prefer to add a polyfill for it since Node 0.12 version will dropped soon. We can move to use `Object.assign` once fastboot supports Node 4.x+

kratiahuja:
@rwjblue This is ready to be reviewed again. FWIW, I also updated a pretty message in this [PR](https://github.com/ember-fastboot/ember-cli-fastboot/pull/280) so that if `najax` is overriden we throw a verbrose error message.

chadhietala:
@kratiahuja Was it intended to add these really large test fixtures? Do these act as a smoke test?

kratiahuja:
@chadhietala yes they act like smoke test.

src/ember-app.js
```diff
@@ -54,23 +55,30 @@ class EmberApp {
...
+    // add any additional user provided variables or override the default globals in the sandbox
+    let globals = {
+      najax: najax,
+      FastBoot: {
+        require: sandboxRequire,
+        config: appConfig
       }
+    };
+    for (let key in addOrOverrideSandboxGlobals) {
+      globals[key] = addOrOverrideSandboxGlobals[key];
```
kellyselden: check Object.prototype.hasOwnProperty here?

kratiahuja: I set this to empty object in index.js if not provided so not needed.

kellyselden: but if it's provided from the outside, you might want to check.

rwjblue:
Can you rebase on top of #75?

kratiahuja:
Sure doing now.

kratiahuja:
@rwjblue rebased and squashed. This is should be ready to merge if everyone is 👍 for this.

kratiahuja:
@rwjblue Rebased and changed the variable name back to `sandboxGlobals` with the comment adding what it can do. 

Once this is merged, I'll work on updating the fastboot website.

PR #95, opened 2016-11-02
add Windows support to development

kellyselden:


stefanpenner:
@kellyselden we should likely also add appveyor (although likely in a later PR)

rwjblue:
Awesome, thanks @kellyselden!

PR #105, opened 2016-12-29
Allow an ordered list of app files to be loaded in sandbox

kratiahuja:
This PR is to address a [usecase](https://github.com/ember-fastboot/ember-cli-fastboot/issues/270) where an app may want to load other assets after the vendor asset is loaded and before the app asset is loaded in the sandbox. Primarily for example, if we want to load i18n translations file, lazy engine assets. Making the `appFiles` an array allows other addons that will run after ember-cli-fastboot to update package.json with the order.

In order to maintain backward compatibility and provide less breakage for production apps, this PR will continue to read app file path from`appFile` manifest (if present). 

Corresponding `ember-cli-fastboot` PR that writes appFiles into an array: https://github.com/ember-fastboot/ember-cli-fastboot/pull/325


test/fixtures/multiple-app-files/assets/fastboot-test.map
```diff
@@ -0,0 +1 @@
...CA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACrCA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;AChCA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACdA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACjBA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACbA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACbA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACbA;AACA;AACA;AACA;AACA;;ACJA;AACA;;ACDA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;ACRA;AACA;AACA;AACA;AACA;AACA;AACA;;ACNA;AACA;AACA;AACA;AACA;AACA;AACA;;ACNA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;;AClDA;AACA;AACA;AACA;AACA;;ACJA;AACA;AACA;AACA;AACA;AACA;AACA;;ACNA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;","file":"fastboot-test.js"}
```
rwjblue: Do we usually check in the `.map` files?

kratiahuja: Yes we do. Seeing it in all fixtures currently.

kratiahuja:
@rwjblue Thanks for the feedback! Here are my responses:

> Should we warn/error if you have both appFile and appFiles?

IMO the build step should not write appFile and appFiles since it will be confusing to the users. 

> Should we add a warning when appFile fallback is used? Maybe this should be in a future version, but it feels like eventually we want only one way to do this.

Yes good idea. It would be better to log it. But in future I would only like to see `appFiles` and support for `appFile` to be dropped.

> Can we do the same treatment for vendorFile? Things like engines (which could totally be made to work by adding them to appFiles before the actual app asset) are really vendor files...

Yes we can. It would be a good idea to make both `vendorFile` and `appFile` as an array. Makes the API more flexible in future too. Will update this PR to make `vendorFile` as an array too. 

> Can you confirm that both appFile and appFiles paths are tested? I believe they are (because all of the existing tests likely use appFile), but I wanted to double check since you are more familiar with what exists at the moment...

Confirmed I tested both paths seperately in  a dummy fastboot app. Also added a test for `appFiles` too. 

@rwjblue I think you brought up a very good point about `vendorFiles`. I'll update this PR to make `vendorFiles` as an array to make the API for both standard and flexible. Will add tests for same too.

kratiahuja:
The Pr is updated per your feedback. @rwjblue could you please review again? If it is good to go, can we please merge it :)

rwjblue:
LGTM!

PR #108, opened 2017-01-09
consistent quotes

kellyselden:


rwjblue:
We ought to land some linting to enforce this stuff. 

Can you create an issue so we can track that? I'd suggest stealing ember-cli's .eslintrc.js...

PR #114, opened 2017-01-15
Introduce schema version in FastBoot

kratiahuja:
## Motivation
`ember-cli-fastboot` publishes a package.json in the dist directory of any ember app that is used by the `fastboot` library to read and load app/vendor files, serve the required html file etc. The `fastboot` micro library is highly dependent on the output format of the package.json. If there is a change in the format of package.json, `fastboot` library needs to be compatible with it. In order to introduce the compatibility, the package.json will contain a `schemaVersion` field. This field will be updated everytime there is an incompatible change from `ember-cli-fastboot`. If apps are using a manifest version that the fastboot library is not aware of (ie manifest version is greater than fastboot latest version), we will throw an error. If the version of fastboot is ahead than the manifest version used by the app we will simply do the backward compatible transformation.

In the long run as we do major version bumps, we will support a limited set of manifest versions.

## Description of changes:
This PR does the following:
- [x] `ember-cli-fastboot` changes are [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/337)
- [x] Hardcodes the latest schema version. Everytime the schema version is updated, this value needs to be updated and a new version minor version needs to be released
- [x] Introduces next schema version: This version depicts the next version that will be introduced. This allows changes future incompatible changes in `fastboot` library to be introduced early but only behind this version flag. The primary purpose of this is to publish `fastboot` library changes early so that `ember-cli-fastboot` integration tests can pass. 
- [x] If the version is not compatible to the latest schema version, a useful error is throw when the fastboot server comes up. This will allow production apps to be aware of such changes in a more informant way.

cc: @rwjblue @tomdale


kratiahuja:
@rwjblue @tomdale : I also took the liberty to drop Node 0.12 from travis.

kratiahuja:
Bump! @rwjblue can you please review this?

src/ember-app.js
```diff
@@ -12,6 +13,11 @@ const debug = require('debug')('fastboot:ember-app');
 const FastBootInfo = require('./fastboot-info');
 const Result = require('./result');
 
+const LATEST_SCHEMA_VERSION = '1';
+// this allows fastboot lib to have unpublished features from ember-cli-fastboot but allows
+// people to run integration tests in ember-cli-fastboot
+const NEXT_SCHEMA_VERSION = '2';
```
tomdale: Perhaps the schema version should be represented as an integer, so `NEXT_SCHEMA_VERSION` can be `LATEST_SCHEMA_VERSION++` and updating the version doesn't require editing two values?

src/ember-app.js
```diff
@@ -344,39 +350,40 @@ class EmberApp {
...
-    if (manifest.appFiles) {
-      debug("reading array of app file paths from manifest");
-      manifest.appFiles.forEach(function(appFile) {
-        appFiles.push(path.join(distPath, appFile));
-      });
-    } else if (manifest.appFile) {
-      // TODO : remove after Fastboot 1.0
-      debug("reading app file path from manifest");
-      appFiles = [path.join(distPath, manifest.appFile)];
+    if (schemaVersion !== LATEST_SCHEMA_VERSION && schemaVersion !== NEXT_SCHEMA_VERSION) {
```
tomdale: I think we should only fail if the version of the schema in the manifest is *newer* than the current schema version. For example, if the manifest version is 57 and the latest the FastBoot server knows about is 13, then there may be information in the manifest we don't understand. In that case, we should fail. But if the version in the manifest is *older*, we know its characteristics and should be able to transform it into the latest manifest schema. That way people can still run older builds while keeping their FastBoot servers up-to-date with the latest version.

src/ember-app.js
kratiahuja: I agree it is cleaner to make it in an integer. Switched to that. However, with the new given scheme of things I don't think we should allow `LATEST_SCHEMA_VERSION++` since it can accidentally cause issues. For example, someone changed the schema version and updated the manifest schema version to 3 while latest version that FastBoot server understands is 2. It will cause breakage.

src/ember-app.js

kratiahuja:
@tomdale Updated per our discussion during the fastboot meeting.

PR #115, opened 2017-01-15
najax upgrade for proper coalesceFindRequests support

izelnakri:
ref: https://github.com/najaxjs/najax/pull/51#issuecomment-272676370


rwjblue:
Thank you!

PR #118, opened 2017-01-17
Fixes #117 - Warn if body/head insertion comments are missing

### LINKED ISSUE #117: Warn if base HTML payload is missing BODY and/or HEAD insertion comments
A non-fastbooted payload is served if the insertion comments are missing from the base HTML. For those using an older version of [ember-cli-html-minifier](https://github.com/gdub22/ember-cli-html-minifier), these comments (and all comments!) are stripped and fastboot fails silently.

The aforementioned addon has added in [fastboot compatibility](https://github.com/gdub22/ember-cli-html-minifier/pull/5), but I think a warning on the fastboot side is prudent as well.

Here's the offending fastboot code: https://github.com/ember-fastboot/fastboot/blob/master/src/result.js#L116-L122

briangonzalez:
Fixes #117

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -3,6 +3,8 @@
 const SimpleDOM = require('simple-dom');
 const HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
 
+const chalk = require('chalk');
+
 /**
  * Represents the rendered result of visiting an Ember app at a particular URL.
  * A `Result` object is returned from calling {@link FastBoot}'s `visit()`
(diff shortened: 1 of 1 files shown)


briangonzalez:
@kratiahuja @danmcclain @arjansingh

src/result.js
```diff
@@ -113,10 +115,20 @@ class Result {
 }
 
 function insertIntoIndexHTML(html, head, body) {
-  html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", body);
+  const headComment = "<!-- EMBER_CLI_FASTBOOT_HEAD -->";
+  const bodyComment = "<!-- EMBER_CLI_FASTBOOT_BODY -->";
+
+  const warnMissing = function (missing) {
```
kratiahuja: Can you use `() => {..}` instead now that we dropped Node 0.12.

src/result.js
```diff
@@ -113,10 +115,20 @@ class Result {
...
+    chalk.red(`Fastboot was not able to find ${missing} in base HTML`);
```
kratiahuja: I think we should throw an error as well. We don't expect developers to use `fastboot` library and call `result.html()` without including `ember-cli-fastboot` addon. They go hand in glove.

briangonzalez:
@kratiahuja There ya go 😉

src/result.js
```diff
@@ -113,10 +115,20 @@ class Result {
 }
 
 function insertIntoIndexHTML(html, head, body) {
-  html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", body);
+  const headComment = "<!-- EMBER_CLI_FASTBOOT_HEAD -->";
+  const bodyComment = "<!-- EMBER_CLI_FASTBOOT_BODY -->";
+
+  const warnMissing = missing => {
```
kratiahuja: Rename this function to something better since you aren't warning now but throwing an error.

rwjblue: agreed

kratiahuja:
Closing this since there is no activity and #149 supersedes this.

PR #119, opened 2017-01-21
Align request protocol with window.location

martletandco:
`window.location.protocol` includes the colon (`https:`) whereas `request.protocol` does not (`https`). As this adds an additional step in getting a consistent protocol value it, introduces a 'gotcha'.

This PR appends a colon to Request's location protocol so that

```
let location = get(this, 'fastboot.isFastBoot') ? get(this, 'fastboot.request') : window.location;
let protocol = get(location, 'protocol');
protocol = protocol.replace(':', '');
```

can just be

```
let location = get(this, 'fastboot.isFastBoot') ? get(this, 'fastboot.request') : window.location;
let protocol = get(location, 'protocol');
```


src/fastboot-request.js
```diff
@@ -4,7 +4,7 @@ var FastBootHeaders = require('./fastboot-headers');
 function FastBootRequest(request, hostWhitelist) {
   this.hostWhitelist = hostWhitelist;
 
-  this.protocol = request.protocol;
+  this.protocol = request.protocol+':';
```
kratiahuja: Spacing for `+` seems a bit odd here.

martletandco: Is the preferred spacing `request.protocol + ':';`?

kratiahuja: I assume so. Better to drop using string concatenation and instead use template liberals in ES6

martletandco: Ok, good call. Have changed this to use a template literal.

martletandco:
Just checking progress on this, are there any objections?

kellyselden:
None from me

kratiahuja:
@martletandco Could you resolve the conflicts? Since there are no strong objections, we should merge this. Also this would be good to get merged before FastBoot 1.0

martletandco:
All up to date now. Didn't see that conflict sorry

kratiahuja:
Thanks @martletandco

PR #128, opened 2017-03-14
Escape '$' as '$$$' due to `String.replace()` special patterns 💰

dtetto:
We found that `$$` in our content ("A$$") was being replaced with `$` ("A$") in production. Believe it or not, this is due to Javascript’s `String.replace()`:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -112,11 +112,19 @@ class Result {
   }
 }
 
+/**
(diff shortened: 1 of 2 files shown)


tomdale:
I had no idea!

PR #129, opened 2017-03-15
Upgrade to najax version 1.0.2

demersus:
The upgrade contains an important fix for catching errors when an
assumed json endpoint returns html.  This can happen when the gateway
times out.

See: https://github.com/najaxjs/najax/pull/56


rwjblue:
Thanks!

PR #131, opened 2017-03-24
Configurable element that shoebox is appended to

rogatty:
I need FastBoot to put the shoebox in the head element, instead of the body. This PR adds an option to override the element using `config/environment.js`:
```
var ENV = {
  fastboot: {
    shoeboxAppendTo: 'head'
  }
}
```

If this look good, I'm going to update [ember-cli-fastboot](https://github.com/ember-fastboot/ember-cli-fastboot/blob/3d1f3f372765df1dad23aa79087e486868b2d59f/lib/broccoli/fastboot-config.js#L168-L179) to add this option to `dist/package.json` instead of accessing `appConfig` directly and will add some tests (and fix the failing ones).

Why I need it in the head? Because we have tracking scripts in the head that have to run before Ember starts and which depend on the data we store in the shoebox. The only other way I see to make it work is to avoid using shoebox and store the data in global variables. It's how we did it before FastBoot and I would prefer to not go this way.

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -39,13 +39,11 @@ class EmberApp {
     this.moduleWhitelist = config.moduleWhitelist;
     this.hostWhitelist = config.hostWhitelist;
     this.appConfig = config.appConfig;
+    this.shoeboxAppendTo = config.shoeboxAppendTo;
 
     if (process.env.APP_CONFIG) {
       this.appConfig = JSON.parse(process.env.APP_CONFIG);
     }
-
-    // FIXME ember-cli-fastboot should add it to dist/package.json
-    this.shoeboxAppendTo = this.appConfig.fastboot.shoeboxAppendTo;
 
     this.html = fs.readFileSync(config.htmlFile, 'utf8');
 
@@ -393,7 +391,8 @@ class EmberApp {
       htmlFile: path.join(distPath, manifest.htmlFile),
       moduleWhitelist: pkg.fastboot.moduleWhitelist,
       hostWhitelist: pkg.fastboot.hostWhitelist,
-      appConfig: pkg.fastboot.appConfig
+      appConfig: pkg.fastboot.appConfig,
+      shoeboxAppendTo: pkg.fastboot.shoeboxAppendTo
     };
   }
 
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -44,6 +44,9 @@ class EmberApp {
       this.appConfig = JSON.parse(process.env.APP_CONFIG);
     }
 
+    // FIXME ember-cli-fastboot should add it to dist/package.json
+    this.shoeboxAppendTo = this.appConfig.fastboot.shoeboxAppendTo;
+
     this.html = fs.readFileSync(config.htmlFile, 'utf8');
 
     this.sandbox = this.buildSandbox(distPath, options.sandbox, options.sandboxGlobals);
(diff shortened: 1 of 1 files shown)


kratiahuja:
Can you not use ember-cli-head for this purpose?

kratiahuja:
My concern is if fastboot introduces streaming in future, your usecase may break.

rogatty:
@kratiahuja by using ember-cli-head, do you mean creating the shoebox divs manually? I.e. using the logic from `createShoebox` to create the same output in the app's `templates/head.hbs`? It _might_ work but it doesn't feel right to duplicate the logic and I worry that it will break as soon as there is any small change in the way the shoebox works.

kratiahuja:
My concern is two fold:

1. The purpose of `shoebox` is for the app to not make the API requests in browser and reuse what the server already used. It's not meant for shoving the data in the `head` and be used by something else. 
2. If you need to explicitly put data (in whatever format you want it doesn't necessarily have to be the format of shoebox) in the head for some other purpose, `ember-cli-head` gives you a flexible API to do so. So why bake the same functionality in this addon?
3. This is more of a long term. If fastboot ever supports streaming, we will have to handle `ember-cli-head` usecase and what your PR is proposing together while if you use `ember-cli-head` we will only have to handle one usecase.

rogatty:
@kratiahuja thank you for spending your time on this. I feel like I didn't explain the issue clearly, so let me try again.

I need to share the same data (specifically route models) between the Ember app and the scripts that are defined in the head element which run before Ember is loaded. It's critical to run these tracking scripts as soon as possible because of the impact on business metrics. I wanted to use the shoebox as it's already the recommended way of passing the data from FastBoot to front-end. I can't do it out of the box because FastBoot puts the shoebox container in the body element and doesn't allow to change the target element.

How can I share the data between the Ember app and the inline head scripts? I don't want to duplicate it by putting route models in both shoebox and global variables (proxying through ember-cli-head). If possible, I would prefer to avoid using global variables at all, using the shoebox provides that.

kratiahuja:
I feel your usecase is out of the scope of fastboot. What happens when your app is not using fastboot? How will you handle then? 

As I mentioned earlier, shoebox is meant in the ember app scope and your proposal most likely will break when streaming is introduced. Did you try using `ember-cli-head`?

rogatty:
Can you explain or point me to documentation which explains, why moving the shoebox to the head element will likely break things when streaming is introduced? If this is really the case then I will have to fallback to inferior solutions like using `ember-cli-head`. As I said, it will mean duplicating the data or resigning from using the shoebox.

The app in question is already tightly coupled to the FastBoot server, so I don't worry about disabling it. On production we run Hapi server which handles all this preparation of data and puts it in the HTML using global variables. Now, we wanted to use Ember instead, without having to maintain two separate codebases, but still need the server-side logic.

kratiahuja:
There is no documentation for streaming since that is not a planned feature currently. I am "assuming" fastboot will introduce streaming at some point. The point of streaming is you send the app and vendor files which are in the body tag soon to the browser while waiting for server side app to finish rendering. With your proposal it would mean we cannot send the app and vendor files unless there is a workaround. 

That being said I am not implying what you have is not a viable solution right now. Streaming was just a long shot which I did not mean to block your work on. In my opinion there already exists a similar solution which is ember-cli-head. I don't think ember-cli-head is an inferior solution. 

Can you fix the tests so that someone else can also look at it and provide feedback?

krisselden:
Closing because discussion has gone stale.

Issue #132, opened 2017-04-06
Unused dependency triggers warning about Regular Expression DoS vulnerability


rogatty:
https://snyk.io/test/github/ember-fastboot/fastboot shows that FastBoot is vulnerable to a ReDoS attack. FastBoot depends on [glob@^4.0.5](https://github.com/ember-fastboot/fastboot/blob/20b687d87d41547e08b35aceff473dc92b7b2570/package.json#L35) which depends on [minimatch@^2.0.1](https://github.com/isaacs/node-glob/blob/v4.5.3/package.json#L22) which is vulnerable. More details about the vulnerability: https://snyk.io/vuln/npm:minimatch:20160620

glob should be updated to at least v7.1.1 where [the issue was fixed](https://github.com/isaacs/node-glob/commit/f0f0872b660d83b1986cd1dd16ec4808fa183adc).

kratiahuja:
Is FastBoot using (requiring) `glob` anywhere? [Seems like it is a dead dependency](https://github.com/ember-fastboot/fastboot/search?utf8=%E2%9C%93&q=glob&type=). But I could be wrong.

rogatty:
You're right, it was introduced in [the first commit](https://github.com/ember-fastboot/fastboot/commit/7829d37194282a30b61455aa6474b4d73745e1ac#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R31) and not used at all. It should be easy fix, then :smile:

danmcclain:
It would be best in the future to follow the [Security guidelines](https://emberjs.com/security/) in the future to minimize exposure. Thanks for the issue still!

PR #133, opened 2017-04-08
Fix deprecated minimatch

Exelord:
Hey,
In this PR I upgraded `glob` to the latest version which depends on `minimatch ^3.0.2`.
Thanks to this, it will fix a deprecation:
> Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "exists-sync": "0.0.3",
     "express": "^4.13.3",
     "express-cluster": "0.0.4",
-    "glob": "^4.0.5",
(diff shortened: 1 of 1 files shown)


rogatty:
Seems like `glob` is not used at all and can be removed from the dependencies, see https://github.com/ember-fastboot/fastboot/issues/132

Exelord:
@rogatty You're right. Changes applied.

rwjblue:
Thank you!

PR #144, opened 2017-06-03
remove `.hasOwnProperty` check for shoebox keys

bekzod:
no need to check for `.hasOwnProperty` if shoebox has empty proto https://github.com/ember-fastboot/ember-cli-fastboot/pull/417

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -430,11 +430,14 @@ function waitForApp(instance) {
  * parse the specific item at the time it is needed instead of everything
  * all at once.
  */
+const hasOwnProperty = Object.prototype.hasOwnProperty; // jshint ignore:line
+
 function createShoebox(doc, fastbootInfo) {
   let shoebox = fastbootInfo.shoebox;
   if (!shoebox) { return; }
 
   for (let key in shoebox) {
+    if (!hasOwnProperty.call(shoebox, key)) { continue; } // TODO: remove this later #144, ember-fastboot/ember-cli-fastboot/pull/417
     let value = shoebox[key];
     let textValue = JSON.stringify(value);
     textValue = escapeJSONString(textValue);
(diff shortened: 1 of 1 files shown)


stefanpenner:
these two pr #417 may need to be the same one?

bekzod:
is possible to make same PR into 2 repos ?

kellyselden:
This is going to be a tricky change, because once one repo gets the change, older versions of the other repo will not work with it. I suggest first checking if hasOwnProperty exists, then use it if it is. Then we can make a similar change in the other repo, release new RCs, then go about removing these blocks.

kratiahuja:
We also have a manifest version to avoid such issues ;)

kellyselden:
Ah I forgot about that!

kratiahuja:
I prefer we punt this change to future. The change is small that it does not warrant for a schema version bump. In future when we do a schema version bump, we can do this change implicitly. thoughts?

stefanpenner:
> thoughts?

Should we keep an open issue somewhere for pending things we want to do for a schema change? That way when we do do one, we have a backlog of changes?

src/ember-app.js
```diff
@@ -435,8 +435,6 @@ function createShoebox(doc, fastbootInfo) {
   if (!shoebox) { return; }
 
   for (let key in shoebox) {
-    if (!shoebox.hasOwnProperty(key)) { continue; }
```
stefanpenner: to future proof this. We could use `Object.hasOwnProperty(showbox, key)` instead?

bekzod: sounds awesome, so it can be merged now :)

bekzod: using `Object.hasOwnProperty(showbox, key)` now, I think now both PR's can be merged

kratiahuja:
Yes I would love for an issue to be opened and we can label it to pick this change implicitly.

kellyselden:
@kratiahuja With the recent changes, this has no consequence with merging. It's the other PR we have to worry about https://github.com/ember-fastboot/ember-cli-fastboot/issues/417.

kratiahuja:
Ok I am good with it then. Didn't see the changes before.

PR #146, opened 2017-06-04
avoid extra lookup when `fastbootInfo` reference already exist

bekzod:
since `fastbootInfo` already instantiated and extra `lookup` does not give any value here

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -215,6 +215,10 @@ class EmberApp {
    * the app instance and then visits the given route and destroys the app instance
    * when the route is finished its render cycle.
    *
+   * Ember apps can manually defer rendering in FastBoot mode if they're waiting
+   * on something async the router doesn't know about. This function fetches
+   * that promise for deferred rendering from the app.
+   *
    * @param {string} path the URL path to render, like `/photos/1`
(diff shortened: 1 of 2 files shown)


src/ember-app.js
```diff
@@ -235,7 +239,7 @@ class EmberApp {
       })
       .then(() => result.instanceBooted = true)
       .then(() => instance.visit(path, bootOptions))
-      .then(() => waitForApp(instance))
+      .then(() => fastbootInfo.deferredPromise)
```
kratiahuja: What if during the `visit`, the instance mutated the `fastbootInfo` object? Can you make sure this isn't a stale copy?

bekzod: even if `fastbootInfo` is mutated it points to same object right, and it is registered in registry as `{ instantiate: false }` which means registry doesn't do anything special it comes out same as it comes in, correct if I am wrong :)

kratiahuja: I am not worried about the registry. I am talking about the mutations during `visit` which is on the instance and not the application. If you can verify the objects are same via a simple test of changing the request property on the `fastbootInfo` in your app, this should be good.

bekzod: added :)

PR #147, opened 2017-06-04
using internal flag `_booted` instead of `instanceBooted`

bekzod:


src/ember-app.js
```diff
@@ -233,7 +233,6 @@ class EmberApp {
 
         return instance.boot(bootOptions);
       })
-      .then(() => result.instanceBooted = true)
```
kratiahuja: Why remove this?

bekzod:
Why to have it ? If app instance already has _booted prop which does same thing, is there reason not using it?

PR #148, opened 2017-06-04
moved app instance destroy logic to `result._destroyAppInstance`

bekzod:
`result._destroyAppInstance` encapsulates all destroy logic which was spread in `visit` method

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -266,7 +266,7 @@ class EmberApp {
     let res = options.response;
     let html = options.html || this.html;
     let disableShoebox = options.disableShoebox || false;
-    let destroyAppInstanceInMs = options.destroyAppInstanceInMs;
+    let destroyAppInstanceInMs = parseInt(options.destroyAppInstanceInMs, 10);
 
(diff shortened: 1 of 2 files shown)


src/ember-app.js
```diff
@@ -310,16 +289,7 @@ class EmberApp {
...
-        if (instance && !result.instanceDestroyed) {
-          result.instanceDestroyed = true;
-          instance.destroy();
-
-          if (destroyAppInstanceTimer) {
-            clearTimeout(destroyAppInstanceTimer);
-          }
-        }
-      });
+      .finally(() => result._destroyAppInstance(destroyAppInstanceInMs));
```
kratiahuja: This is not the desired behavior that we want. We want to start a timer before the `visit` request starts in order to make sure we destroy the instance forcefully if the request never completes.

With your refactor, we will start the timer after the `visit` request is completed which might not be of any use.

bekzod: got it, thanks for clarify I missed that part, let me give it another shot

Issue #76, opened 2016-08-05
ApplicationInstance Boot Errors are being swallowed


arjansingh:
I ran into a particularly vicious error, where a stray `window` reference caused the `ApplicationInstance` to fail booting:

```
  _bootPromise:
   Promise {
     _id: 14,
     _label: undefined,
     _state: 2,
     _result: [TypeError: window.addEventListener is not a function],
     _subscribers: [],
     _onError: null } }
```

The failure was silent and I only detected it when I logged `this.instance` inside of [`src/result.js`](https://github.com/ember-fastboot/fastboot/blob/master/src/result.js#L56-L67).

We should be logging these events (not swallowing them silently). 

Of course, I'll open a PR but I want to know if `result._finalize` is the place to do the error handling. Suggestions welcome.

arjansingh:
Good news (and bad news)! It's not the fastboot middleware. I am able to get this repo to error out locally using `ember-cli-fastboot`. 

I did some preliminary digging and it looks like the boot errors are localized to the instance of [fastboot-app-server](https://github.com/ember-fastboot/fastboot-app-server) that I'm using.

I'll keep digging in that repo to see if I can reach some actual lines of code.

arjansingh:
@danmcclain @tomdale 

### Resilient mode is the cause
I was poking around today and found the ultimate source of this issue's problems.

If you have [resilient mode turned on](https://github.com/ember-fastboot/fastboot/blob/master/src/index.js#L70-L77) we just swallow the error instead of throwing it. 

`fastboot-app-sever` has [resilient mode on](https://github.com/ember-fastboot/fastboot-app-server/blob/500c29af6f08715a7b637a378d079c1b8cab744e/src/worker.js#L59-L62).

### Potential solutions
1. Modify the error handling in this project.
2. Modify the error handling in `fastboot-app-server` and add documentation to warn about resilient mode, specifically about how end-users are responsible for manually handling result errors if it is enabled.

Thoughts?

tomdale:
This seems like it's working as intended? FastBoot App Server is designed for production, where you typically want to proceed with client-side rendering if server-side rendering fails. It should be caught in development, where `resilient` should be turned off.

Maybe we should add a note recommending running a staging/dev environment with it off?

arjansingh:
We still want logging for these errors `fastboot-app-server` because they can still happen in production. Someone could download some bad Ember application code (e.g, a stray `window` reference) on their server and break FastBoot rendering.  

[...]

PR #149, opened 2017-06-04
throw when insert comments are not replaced

bekzod:
https://github.com/ember-fastboot/fastboot/issues/76#issuecomment-272971321

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -43,7 +43,7 @@ class Result {
       }
     }
 
-    return Promise.resolve(insertIntoIndexHTML(this._html, this._head, this._body));
+    return insertIntoIndexHTML(this._html, this._head, this._body);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bekzod:
this also makes `escapeForStringReplace` redundant 🎊

src/result.js
```diff
@@ -112,22 +112,34 @@ class Result {
...
+
+  if (body) {
+    let isBodyReplaced = false;
+    html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", function() {
+      isBodyReplaced = true;
+      return body;
+    });
+
+    if (!isBodyReplaced) {
+      return Promise.reject(new Error("`<!-- EMBER_CLI_FASTBOOT_BODY -->` `insert comment` is missing in your index.html"));
```
kratiahuja: When will this comment be missing? `fastboot` cannot work with `ember-cli-fastboot` which inserts these comments.

src/result.js
```diff
@@ -112,22 +112,34 @@ class Result {
...
-}
-
 function insertIntoIndexHTML(html, head, body) {
-  html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", escapeForStringReplace(body));
+  if (!html) { return Promise.resolve(html); }
+
+  if (body) {
+    let isBodyReplaced = false;
+    html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", function() {
+      isBodyReplaced = true;
```
kratiahuja: When will they not be replaced? Is there a usecase you had in mind?

kratiahuja:
There is also a similar PR to warn if the comments are missing: #https://github.com/ember-fastboot/fastboot/pull/118 . I would prefer we warn and still send the index.html.

src/result.js
bekzod: what do you mean ? if `<!-- EMBER_CLI_FASTBOOT_BODY -->` missing `isBodyReplaced` will not be `true`, therefore will throw error

bekzod:
I can change throw to warn, #118 doesn't do it efficiently it has extra `indexOf`s

bekzod:
#118 also does not get rid of `escapeForStringReplace`

bekzod:
so this should warning not assertion ?

kratiahuja:
Nevermind, I think error is fine. The PR's heading is confusing.

src/result.js
```diff
@@ -112,22 +112,34 @@ class Result {
...
   if (head) {
-    html = html.replace("<!-- EMBER_CLI_FASTBOOT_HEAD -->", escapeForStringReplace(head));
+    let isHeadReplaced = false;
+    html = html.replace("<!-- EMBER_CLI_FASTBOOT_HEAD -->", function() {
+      isHeadReplaced = true;
+      return head;
+    });
+
+    if (!isHeadReplaced) {
+      return Promise.reject(new Error("`<!-- EMBER_CLI_FASTBOOT_HEAD -->` `insert comment` is missing in your index.html"));
```
kratiahuja: Let's make the error message a bit more descriptive. Example: `Fastboot was not able to find <!--EMBER_CLI_FASTBOTT_HEAD--> in base HTML. It could not replace the contents.`

src/result.js
kratiahuja: You can wrap the error message in its function to avoid duplicating it in two places.

src/result.js
bekzod: sure

src/result.js
bekzod: ok let me see

src/result.js
kratiahuja: I was more curious to understand the motivation behind the change. Did you hit this usecase where the comments weren't present?

bekzod: No just wanted get rid of `escapeForStringReplace` because it was doing extra `.replace` calls, and also come across this comment https://github.com/ember-fastboot/fastboot/issues/76#issuecomment-272971321 so just did two things in one shot :P

src/result.js
bekzod: done

src/result.js
bekzod: done

kratiahuja:
Thank you @bekzod

PR #150, opened 2017-06-04
fixed tests, now they are broken

bekzod:
mocha has behavior of setting test as passing if `promise` returned inside `it(....)` passing,
but if you have assertion in `.catch` it might skip assertion because returned promise resolved,
to avoid that you need finish assertion with `done`

because of it I think some of errors have been missed in master @kratiahuja @krisselden


kratiahuja:
I thought mocha has the same behavior with returning promise or using `done`: https://mochajs.org/#working-with-promises

Also not all tests are failing when using `done` in a `catch`.

bekzod:
Yes of course, i wrote some therefore :P, I think only one failing other one i think just jshint error

kratiahuja:
There are two failing. One is a `jshint`.

bekzod:
@kratiahuja I am look at those error, will make PR if I fix them thanks

test/fastboot-test.js
```diff
@@ -101,16 +101,15 @@ describe("FastBoot", function() {
       });
   });
 
-  it("can forcefully destroy the app instance using destroyAppInstanceInMs", function() {
+  it("can forcefully destroy the app instance using destroyAppInstanceInMs", function(done) {
```
stefanpenner: why `done`?

bekzod: because otherwise it is skipping assertion in catch, you can try yourself

kratiahuja: I a bit surprised with the mocha behavior if it is right.  I thought returning promise or `done` was the same in mocha.

kratiahuja: I think this should be fixed in mocha upstream as well.

bekzod: I think since you are returning `resolved` promise it is behaving correctly, it either needs rejections inside `then` or it needs to use `done`
> I think this should be fixed in mocha upstream as well.

bekzod:
hurray tests fixed 🎊

test/fastboot-test.js
```diff
@@ -101,16 +101,26 @@ describe("FastBoot", function() {
...
     var fastboot = new FastBoot({
       distPath: fixture('basic-app')
     });
 
-    return fastboot.visit('/', {
-      destroyAppInstanceInMs: 5
-    })
+    // delaying `visitRoute` to forcefully destroy app instance
+    let originalVisitRoute = fastboot._app.visitRoute;
```
kratiahuja: Instead of mutating the private API in the test to force a condition, please create a fixture that does a `setTimeout` in the model hook which implicitly will delay the `visit` resolution.

It seems `basic-app` does not do this currently.

bekzod: added `long-resolving-app` with application route that has model which resolves in 10 seconds, that revealed 2 problems:
* timed destroy could be fired too early before app instance created (so ended up not destroying app)
* when app destroyed before `instance.visit` is called, it was throwing error for setting property for destroyed object (which is still thrown but it is superseded by `App instance was forcefully destroyed ...`).

both of this issues handled now

src/ember-app.js
```diff
@@ -226,21 +226,31 @@ class EmberApp {
...
-    let instance;
-
+  visitRoute(path, fastbootInfo, bootOptions, result, destroyAppInstanceInMs) {
     return this.buildAppInstance()
-      .then(appInstance => {
-        instance = appInstance;
+      .then(instance => {
         result.instance = instance;
-        registerFastBootInfo(fastbootInfo, instance);
+        return new Promise(function(resolve, reject) {
```
stefanpenner: `=>` ?

src/ember-app.js
```diff
@@ -226,21 +226,31 @@ class EmberApp {
...
+          let timer;
+          if (destroyAppInstanceInMs > 0) {
+            // start a timer to destroy the appInstance forcefully in the given ms.
+            // This is a failure mechanism so that node process doesn't get wedged if the `visit` never completes.
+            timer = setTimeout(()=> {
```
stefanpenner: funky whitespace between `()` and `=>`

test/fastboot-test.js
stefanpenner: In general I find https://github.com/domenic/chai-as-promised clears all this up nicely.

bekzod: will cleanup with next PR, using it :)

bekzod:
good to merge ? :P

test/fastboot-test.js
```diff
@@ -101,16 +101,15 @@ describe("FastBoot", function() {
...
+      distPath: fixture('long-resolving-app')
     });
 
-    return fastboot.visit('/', {
-      destroyAppInstanceInMs: 5
-    })
+    fastboot.visit('/', { destroyAppInstanceInMs: 5 })
       .catch((e) => {
-        expect(e.message).to.equal('App instance was forcefully destroyed in 5ms');
+        expect(e.message).to.equal('Fastboot forcefully destroyed App instance in 5ms');
```
stefanpenner: if this expect fails, I believe an assertion is thrown preventing `done` from being invoked. This may result in annoying test issues on failure.

bekzod: test will just timeout, current timeout 1000ms I think, so it will still fail,
but if I return whole promsie `return fastboot.visit('/', ` latest mocha will complain that 
we are `returning promise` and using `done` at the same time, thoughts ?

kratiahuja: So why can't we just return the promise and move the assertion in the reject function?

bekzod: @kratiahuja because of this https://github.com/ember-fastboot/fastboot/pull/150#issue-233426083 , the whole point of this PR was not to that, because of mentioned problems :)

bekzod:
this probably fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/209 too

kratiahuja:
No those are different issues. The new config was introduced later.

PR #151, opened 2017-06-06
refactored `insertIntoIndexHTML`

bekzod:


src/result.js
```diff
@@ -127,29 +127,25 @@ function missingTag(tag) {
 
 function insertIntoIndexHTML(html, head, body) {
   if (!html) { return Promise.resolve(html); }
+  let isBodyReplaced = false;
+  let isHeadReplaced = false;
 
-  if (body) {
-    let isBodyReplaced = false;
-    html = html.replace("<!-- EMBER_CLI_FASTBOOT_BODY -->", function() {
+  html = html.replace(/<\!-- EMBER_CLI_FASTBOOT_(HEAD|BODY) -->/g, function(match, p1) {
```
kratiahuja: My two cents:

I really find the changes convoluted and the `head` and `body` fork is at two places instead of one - making readability hard. I am not convinced this is improving readability if that is what the motivation of the changes was.

bekzod: motivation is was improve performance :P,  one `.replace` is better than `.replace`s, simple benchmark I used https://gist.github.com/bekzod/f08628db3cbbbd678bbc8f72e4b01001 , readability-wise I would say it is same level as before at least it uses less lines of code

kratiahuja: By how much was the performance improved?

bekzod:
```
before change: 370ms
after change: 193ms
```
so 50%

bekzod: at least on my machine (mac os node v6.8.1)

kratiahuja: can we use a better variable than `p1`?

src/result.js
```diff
@@ -127,29 +127,25 @@ function missingTag(tag) {
...
-    let isHeadReplaced = false;
-    html = html.replace("<!-- EMBER_CLI_FASTBOOT_HEAD -->", function() {
-      isHeadReplaced = true;
-      return head;
-    });
-
-    if (!isHeadReplaced) {
-      return missingTag('<!--EMBER_CLI_FASTBOTT_HEAD-->');
-    }
+  if (head && !isHeadReplaced) {
```
kratiahuja: Why check for head and body here? Seems unecessary since it defaults to empty string.

bekzod: https://github.com/bekzod/fastboot/blob/45087528437aa0d1c2dc620b9a45898a8063ce1a/src/result.js#L37-L39 because sometimes missing head insert comment tag is still valid, the same logic was used before, if I remove head bunch of test are broken

src/result.js
bekzod: done

src/result.js
kratiahuja: But `body` will always be present.

PR #153, opened 2017-06-10
`var` => `const`

bekzod:


src/fastboot-request.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
-var cookie = require('cookie');
-var FastBootHeaders = require('./fastboot-headers');
+const cookie = require('cookie');
```
kratiahuja: This seems unused. Can you double check?

bekzod: checked its used in line 51, all good

PR #154, opened 2017-06-14
Assert require sandbox object in EmberApp constructor

tsubomii:
Due to the API change in this commit  https://github.com/ember-fastboot/fastboot/commit/b155d0f9cbe7d2079b65a7819921394975afcc8f, Sandbox object become required argument to construct EmberApp. Add obvious assertion message for easy debugging.

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -26,10 +26,14 @@ class EmberApp {
    * Create a new EmberApp.
    * @param {Object} options
    * @param {string} options.distPath - path to the built Ember application
-   * @param {Sandbox} [options.sandbox=VMSandbox] - sandbox to use
+   * @param {Sandbox} options.sandbox=VMSandbox - sandbox to use
    * @param {Object} [options.sandboxGlobals] - sandbox variables that can be added or used for overrides in the sandbox.
    */
   constructor(options) {
+    if (typeof options.sandbox === 'undefined' || !(options.sandbox instanceof Object)) {
+      throw new Error('Sandbox object is required option to construct EmberApp object');
+    }
+
     let distPath = path.resolve(options.distPath);
     let config = this.readPackageJSON(distPath);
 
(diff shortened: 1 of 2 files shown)


src/ember-app.js
```diff
@@ -26,10 +26,14 @@ class EmberApp {
    * Create a new EmberApp.
    * @param {Object} options
    * @param {string} options.distPath - path to the built Ember application
-   * @param {Sandbox} [options.sandbox=VMSandbox] - sandbox to use
+   * @param {Sandbox} options.sandbox=VMSandbox - sandbox to use
    * @param {Object} [options.sandboxGlobals] - sandbox variables that can be added or used for overrides in the sandbox.
    */
   constructor(options) {
+    if (typeof options.sandbox === 'undefined' || !(options.sandbox instanceof Object)) {
```
kratiahuja: I don't think this is right. We default to `vm-sandbox` if the sandbox class is not provided. See here: https://github.com/ember-fastboot/fastboot/blob/master/src/index.js#L48

So this check is not needed.

tsubomii: That's only in the case when user directly use private API EmberApp, which we currently are. I am working on removing those private API access.

kratiahuja: I don't think any other consumer is doing it such a way and neither that is the recommendation.

stefanpenner: When this assertion fires, it seems like a good teachable moment. Is there something concise we can say, to get people using public API? Or if its a tad longer, maybe a short markdown file can be provided to docs/ directory?


If that providing additional details here is redundant, should we leave a helpful assertion or is that harmful?

kratiahuja: Yeah we can point to the README but it isn't updated. My bad. I'll do it today.

kratiahuja: THe readme is fixed with this PR: https://github.com/ember-fastboot/fastboot/pull/156 . We should point to the readme as a helpful assertion message.

krisselden:
How code is evaluated is internal 3.0 so this no longer makes sense to expose an object that has eval() and run() methods how they were.

Need to understand the use case so we can talk about new API to address it because this API as is is not going to come back in 3.0.

PR #157, opened 2017-06-22
Dependencies cleanup

bobisjan:
What's changed:

- `express` is required only for testing,
- `express-cluster` and `miniminst` are not required at all.


stefanpenner:
@bobisjan good catch, thanks for the cleanup!

PR #164, opened 2017-08-27
Extend FastBoot.config to read other provided config as well.

### LINKED ISSUE #163: Extend `FastBoot.config` API for engines
Parity with https://github.com/ember-fastboot/ember-cli-fastboot/issues/513

cc: @rwjblue

kratiahuja:
Fixes #163 . This will unblock engines support in FastBoot.

# Motivation

Today by default FastBoot provides the ability to read config based info in Node for host applications only. Other infrastructure like Ember Engines may provide their own configuration per bundle. In order to support this, we want to extend `FastBoot.config` to have a `key` as an input which determines which config to return a JSON blob.

# Things done in this PR

- [x] Bumped the supported manifest version of `fastboot` to 3 since the format has changed
- [x] App config by default will be read from `app-name/config/environment
- [x] Backward compatibility for older schema versions
- [x] Added relevant tests

# FastBoot addon

Corresponding FastBoot addon PR: https://github.com/ember-fastboot/ember-cli-fastboot/pull/515

cc: @rwjblue


src/ember-app.js
```diff
@@ -37,10 +37,15 @@ class EmberApp {
...
+    this.config = config.config;
+    this.appName = config.appName;
 
     if (process.env.APP_CONFIG) {
-      this.appConfig = JSON.parse(process.env.APP_CONFIG);
+      let appConfig = JSON.parse(process.env.APP_CONFIG);
+      let appConfigKey = `${this.appName}/config/environment`;
+      if (!appConfig.hasOwnProperty(appConfigKey)) {
+        this.config[appConfigKey] = appConfig;
+      }
```
rwjblue: We should probably create another environment variable to allow the entire config to be replaced, thoughts?

src/ember-app.js
```diff
@@ -60,17 +65,27 @@ class EmberApp {
    */
   buildSandbox(distPath, sandboxClass, sandboxGlobals) {
     let sandboxRequire = this.buildWhitelistedRequire(this.moduleWhitelist, distPath);
-    let config = this.appConfig;
-    function appConfig() {
-      return { default: config };
+    const config = this.config;
+    const appName = this.appName;
+    function fastbootConfig(key) {
+      if (!key && config) {
```
rwjblue: I don't think this guard is needed, defaulting the `key` here when not provided to `appname/config/environment` (regardless if `config` is available) seems totally fine, no?

src/ember-app.js
```diff
@@ -37,10 +37,15 @@ class EmberApp {
...
     this.moduleWhitelist = config.moduleWhitelist;
     this.hostWhitelist = config.hostWhitelist;
-    this.appConfig = config.appConfig;
+    this.config = config.config;
+    this.appName = config.appName;
 
     if (process.env.APP_CONFIG) {
-      this.appConfig = JSON.parse(process.env.APP_CONFIG);
+      let appConfig = JSON.parse(process.env.APP_CONFIG);
+      let appConfigKey = `${this.appName}/config/environment`;
```
rwjblue: Should the keys be `${appName}/config/environment` or just `appName`?  I'd generally prefer the config keys to be "opaque" (so either is fine), but I'd lean towards making the key in this object just `appName` here (though I don't feel _super_ strongly).

src/fastboot-schema-versions.js
```diff
@@ -7,9 +7,10 @@
  * should be added in fastboot lib) everytime fastboot addon schema version is bumped.
  */
 const FastBootSchemaVersions = {
-  'latest': 2, // latest schema version supported by fastboot library
+  'latest': 3, // latest schema version supported by fastboot library
   'base': 1, // first schema version supported by fastboot library
-  'manifestFileArrays': 2 // schema version when app and vendor in manifest supported an array of files
+  'manifestFileArrays': 2, // schema version when app and vendor in manifest supported an array of files
+  'configExtension': 3 // schema version when FastBoot.config can read artibitary indexed config
```
rwjblue: typo here, `artibitary` -> `arbitrary`

src/ember-app.js
kratiahuja: I am fine with anything. The reason I went with `appName/config/environment` since it is easier to map to the AMD module name.

src/ember-app.js
kratiahuja: What's the use case for that?

src/ember-app.js
kratiahuja: Yeah I might have forgotten to clean this up..

src/ember-app.js
rwjblue: Well, the same as the use case for `process.env.APP_CONFIG` in the first place.  To allow customizing the configuration without requiring a recompile.

kratiahuja: Oho yes, I agree. Will add that.

rwjblue:
@kratiahuja - Ping?

kratiahuja:
@rwjblue updated per feedback. Super sorry for the delay here!

PR #165, opened 2017-09-11
[WIP] Allow AMP File to be used based on toggle

devotox:
@kratiahuja 

I was thinking something like this or possibly allow for a map of different HTML files based on a key which could be the path or a regex but this would mean the readFileSync would have to be done on every visit and would not be performant

Corresponding PR In ember-cli-fastboot
https://github.com/ember-fastboot/ember-cli-fastboot/pull/525

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -44,6 +44,7 @@ class EmberApp {
     }
 
     this.html = fs.readFileSync(config.htmlFile, 'utf8');
+    this.amp = this.htmlFile !== this.ampFile ? fs.readFileSync(config.ampFile, 'utf8') : this.html;
 
     this.sandbox = this.buildSandbox(distPath, options.sandbox, options.sandboxGlobals);
     this.app = this.retrieveSandboxedApp();
@@ -267,8 +268,8 @@ class EmberApp {
   visit(path, options) {
     let req = options.request;
     let res = options.response;
-    let html = options.html || this.html;
     let disableShoebox = options.disableShoebox || false;
+     let html = options.html || options.ampMode ? this.amp : this.html;
     let destroyAppInstanceInMs = parseInt(options.destroyAppInstanceInMs, 10);
 
     let shouldRender = (options.shouldRender !== undefined) ? options.shouldRender : true;
@@ -370,6 +371,7 @@ class EmberApp {
     return {
       appFiles: appFiles,
       vendorFiles: vendorFiles,
+      ampFile: path.join(distPath, manifest.ampFile),
       htmlFile: path.join(distPath, manifest.htmlFile),
       moduleWhitelist: pkg.fastboot.moduleWhitelist,
       hostWhitelist: pkg.fastboot.hostWhitelist,
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -44,7 +44,7 @@ class EmberApp {
     }
 
     this.html = fs.readFileSync(config.htmlFile, 'utf8');
-    this.amp = this.htmlFile !== this.ampFile ? fs.readFileSync(config.ampFile, 'utf8') : this.html;
+    this.amp = config.htmlFile !== config.ampFile ? fs.readFileSync(config.ampFile, 'utf8') : this.html;
 
     this.sandbox = this.buildSandbox(distPath, options.sandbox, options.sandboxGlobals);
     this.app = this.retrieveSandboxedApp();
--- a/src/index.js
+++ b/src/index.js
@@ -76,6 +76,13 @@ class FastBoot {
       resilient = this.resilient;
     }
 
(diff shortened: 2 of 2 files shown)


kratiahuja:
I don't think this is the approach I would like to take.

Why can't this be done in a `beforeMiddleware` hook of fastboot-app-server? You could define two package.json files and create two fastboot instances and decide to call `visit` based on the query param. This way it is baked in as an opt-in via a middleware rather than in the library itself.

devotox:
I don't see how this is possible outside the library without creating your own HTTP Server and fastboot instance. Just the middleware hook alone does not give you access to the visit call and also the package.json file is created by ember CLI. Could you elaborate how I could create 2 of these and also how I can access the visit call

kratiahuja:
YOu are right that you would need to create your own HTTP server but I would rather go down that path since this is something that can be easily built and customized outside of fastboot itself. In general I am -1 to this approach of adding specific things when they can be built using the existing infrastructure.

devotox:
Thats understandable but then what about a new hook in the fastboot server
that allows us to then customize the visit function and then in that way we
can still have the fully production ready server. Would you be open to that
option something like the beforeMiddleware maybe call it visit middleware?

kratiahuja:
Possibly yes but it would be great if you could hash out the API for that before doing the actual coding work.

devotox:
```javascript
   * @param {string} path the URL path to render, like `/photos/1`
   * @returns {string} html the html string to be passed to the visit function
onVisit(path) {
 // Return html string or any falsey value will use the index.html file
}

// fastboot/src/index.js

visit(path, options) {
  ....
   	// Must return an html string
	// with fs.readFileSync(<path>, 'utf8')
	if (this.onVisit) {
	   let htmlString = this.onVisit(path);
           if (htmlString) {
             options.html = htmlString
          }
	}
   ....
}

```

Another possibility I considered was just having a visitOptions object which gets passed all the way down and can then just extend the current options with a simple Object.assign

kratiahuja:
I am not sure I follow this API at all. Why do you have to update the `fastboot` library? As mentioned, I would rather have us build a hook in the fastboot-app-server. I am going to close this PR as mentioned. Feel free to send a gist of what you have in mind in terms of the API. But TL;DR is:

FastBoot already has an API to return you DOM contents instead of serialized index.html. You could use that instead and append to AMP via a middleware hook.

devotox:
Haaaa I misunderstood you initially I thought you wanted me to make a middleware type hook that went down to the fastboot library. I shoula actually make the changes in fastboot-express-middleware and fastboot-app-serverHere is a gist of the possible changes. I also added ember-cli-fastboot to allow the same option to be passed in from the app confighttps://gist.github.com/devotox/5c9e0a5684ce321f7d44abb2027b4024Best regards,
Devonte Emokpae
Software Solutions Architect
Zen__
"Meeting all your software solutions needs"
M: +44 7447 903 346
W: Stack Overflow
L: LinkedIn
E: me@devonte.io
          
            
            On Sep 14 2017, at 1:38 am, Krati Ahuja <notifications@github.com> wrote:
            
            
              I am not sure I follow this API at all. Why do you have to update the fastboot library? As mentioned, I would rather have us build a hook in the fastboot-app-server. I am going to close this PR as mentioned. Feel free to send a gist of what you have in mind in terms of the API. But TL;DR is:
FastBoot already has an API to return you DOM contents instead of serialized index.html. You could use that instead and append to AMP via a middleware hook.

kratiahuja:
My apologies to not getitng back here. 

Yeah you would need changes in fastboot-express-middleware and fastboot-app-server as well. On rethinking, one idea that you could do is have `visit` function in fastboot provide the optional html file to write in (more like a runtime config) and if that is present, fastboot appends the serialized template in there. Then in fastboot-app-server, you would need pass that option in visit API. If the optional file is not provided, fastboot defaults back to the htmlFile provided in package.json. Would something like that work for you?

devotox:
Wouldnt that be what the gist above is doing the onVisit API returns some html or not and if nothing it falls back to using the htmlFile option which defaults to index.html if nothing passed in

kratiahuja:
Oho I probably didn't follow onVisit. I think you may not need onVisit if you Enhance the visit API to take optional html file.

devotox:
But you still need a way to have the optional file be chosen based on the
path

kratiahuja:
Yes but you could write it in `dist` and read from there like the default `htmlFile`. My point is you wouldn't need to put in the package.json. 

I think it might be easier if one of just spikes this. Let me find some cycles and spike this to explain better what I have in mind that might work in your use case.

devotox:
Alright sounds good. At the moment I'm using fastboot directly and that is
ok for now. We can flesh out the correct way to implement this and would be
happy to work on it

PR #176, opened 2017-11-22
add API for response chunks

### LINKED ISSUE #174: Allow chunked transfer encoding
I've been looking into ways to easily get chunked transfer encoding with FastBoot, the main motivation being that I'd like to have a mechanism where big shoeboxes don't unnecessarily delay initial rendering of the page.

I believe a really easy way to get this should be to send the response in 3 chunks:

* everything until `</head>`
* everything until the first shoebox if present (or the complete rest of the document if not)
* the shoebox(es) if present and the final `</body></html>`

Of course being able to send chunks of the document as it is rendered on the server side would be ideal but I guess that's a bigger change in the FastBoot internals so the above solution might be a good start?

I looked at the code and I guess it should be straight forward to add a `chunks()` method (or so) to https://github.com/ember-fastboot/fastboot/blob/master/src/result.js and modify https://github.com/ember-fastboot/fastboot-express-middleware so that it writes each of the chunks instead of the complete HTML at once but I wanted to get some feedback before creating a PR.

marcoow:
This adds a new method `chunks` to the `Result` class that returns the HTML in chunks:

* if there are no shoeboxes in the document, it returns 2 chunks - one for the header and one for the body
* if there are shoeboxes in the document, it returns n chunks - one for the header, one for the body (except the shoeboxes that are at the bottom of the body) and one for each shoebox

The idea behind this is that this could be used in e.g. https://github.com/ember-fastboot/fastboot-express-middleware to send a streaming response so that big shoeboxes don't delay the whole document.

closes #174

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -3,6 +3,9 @@
 const SimpleDOM = require('simple-dom');
 const HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
 
+const SHOEBOX_TAG_PATTERN = '<script type="fastboot/shoebox"';
+const HTML_HEAD_REGEX = /^([\s\S]*<\/head>)([\s\S]*)/;
+
 /**
  * Represents the rendered result of visiting an Ember app at a particular URL.
  * A `Result` object is returned from calling {@link FastBoot}'s `visit()`
@@ -58,17 +61,15 @@ class Result {
    */
   chunks() {
     return insertIntoIndexHTML(this._html, this._head, this._body, this._bodyAttributes).then((html) => {
-      let [, head, body] = html.match(/^([\s\S]*<\/head>)([\s\S]*)/);
+      let [, head, body] = html.match(HTML_HEAD_REGEX);
       let chunks = [head];
-      if (/<script type="fastboot\/shoebox"/.test(body)) {
-        let [strippedBody, shoeboxes] = body.match(/^([\s\S]*?)(<script type="fastboot\/shoebox"[\s\S]*)/).splice(1);
-        chunks.push(strippedBody);
-        shoeboxes.split(/<script type="fastboot\/shoebox"/).splice(1).forEach((shoebox) => {
-          chunks.push(`<script type="fastboot/shoebox"${shoebox}`);
-        });
-      } else {
-        chunks.push(body);
-      }
+
+      let [plainBody, ...shoeboxes] = body.split(SHOEBOX_TAG_PATTERN);
+      chunks.push(plainBody);
+      shoeboxes.forEach((shoebox) => {
+        chunks.push(`${SHOEBOX_TAG_PATTERN}${shoebox}`);
+      });
+
       return chunks;
     });
   }
--- a/src/result.js
+++ b/src/result.js
@@ -46,6 +46,34 @@ class Result {
   }
 
   /**
+   * Returns the HTML representation of the rendered route, inserted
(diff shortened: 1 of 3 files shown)


src/result.js
```diff
@@ -3,6 +3,9 @@
 const SimpleDOM = require('simple-dom');
 const HTMLSerializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
 
+const SHOEBOX_TAG_PATTERN = '<script type="fastboot/shoebox"';
+const HTML_HEAD_REGEX = /^([\s\S]*<\/head>)([\s\S]*)/;
```
SergeAstapov: @marcoow it does not support body attributes recently added in #171

marcoow: I'm not sure what you mean - this RegExp does not make any assumptions on how the `<body>` looks and works perfectly fine with a tag like `<body data-foo="my value">`.

SergeAstapov: @marcoow ah sorry for confusion, my bad - it says `head>` and I was reading `body>` 🤦

marcoow:
@kratiahuja, @rwjblue: do you have a minute to review?

kratiahuja:
@marcoow: sorry I have been travelling. I can review it early next week if that isn’t late.

marcoow:
Thanks @kratiahuja! I'm sure your busy so whenever you find the time is great but we'd like to use this from a release version ;)

src/result.js
```diff
@@ -45,6 +48,45 @@ class Result {
...
+   * The first chunk contains the document's head, the second contains the body
+   * until just before the shoebox tags (if there are any) and the last chunk
+   * contains the shoebox tags and the closing `body` tag. If there are no
+   * shoebox tags, there are only 2 chunks and the second one contains the
+   * complete document body, including the closing `body` tag.
+   *
+   * @returns {Promise<Array<String>>} the application's DOM serialized to HTML, split into chunks
+   */
+  chunks() {
+    return insertIntoIndexHTML(this._html, this._head, this._body, this._bodyAttributes).then((html) => {
```
kratiahuja: A minor optimization that can be done here is:

Instead of replacing the entire index.html, you may want to think about only replacing the chunk lazily. This will enable sending the head faster. Can be done at a later point too. I think with streaming you want to make sure you do an early flush as soon as possible.

kratiahuja:
@marcoow I'll release a new version tonight.

marcoow:
@kratiahuja: awesome - thanks!

src/result.js
marcoow: Yeah, ideally we could stream the head as soon as it is rendered, even while the rest of the document is still being rendered. My understanding is that that is currently not possible though as we have to wait until `visit` resolves at which point the complete document has been rendered already? Also I guess in order for that to work we'd have to make this a generator function rather than returning a promise?

Or maybe you're referring to something else that I don't get :)

kratiahuja:
@marcoow `fastboot@1.1.2` is released!

src/result.js
kratiahuja: Yeah I wasn't refering to waiting for `visit`. But right now we parsing the entire `index.html` and replacing body as well when we only want to send the head first.

marcoow: Yeah, good point. But I think we'd have to turn this into a generator function then though to be able to give control back to the caller once the first chunk has been calculated. That would be a breaking change though…

PR #180, opened 2018-01-09
Allow html attributes

SergeAstapov:
The SimpleDOM document that gets created during fastboot rendering has a html tag, and it's possible to set attributes on it. For example, assuming getDOM from [ember-wormhole](https://github.com/yapplabs/ember-wormhole/blob/6992bb2a9fced9183cdb426a931722fb1fefba8a/addon/utils/dom.js#L45-L63):

```
Component.extend({
  init() {
    this._super(..arguments);
    getDOM(this).documentElement.setAttribute('data-foo', 'hello')
  }
})
```

But this doesn't actually cause any output in fastboot, because the user's own <html> tag from index.html is always used instead.

This PR allows attributes set on the SimpleDOM html to be copied into the output.

Credit to @ef4 for inspirational PR #171.

Closes ember-fastboot/ember-cli-fastboot#331


SergeAstapov:
Hi @kratiahuja! Do you mind to take a look once get a chance? Do you feel this is right way to approach the problem?

bobisjan:
This would be a great addition for using lang attribute on HTML element 👍.

rwjblue:
Sorry for the delay here! Merged and released as 1.2.0...

PR #181, opened 2018-01-17
Add markers before and after the body to be able to remove rootless apps

cibernox:
@rwjblue @kratiahuja Would you consider this breaking?

It must be released along with a new version of ember-cli-fastboot that knows how to remove DOM elements between those new two markers.

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -183,7 +183,7 @@ function insertIntoIndexHTML(html, head, body, bodyAttributes) {
       return head;
     } else if (tag === 'BODY' && body && !isBodyReplaced) {
       isBodyReplaced = true;
-      return body;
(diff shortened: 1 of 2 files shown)


rwjblue:
Doesn’t seem breaking to me. If used with current ember-cli-fastboot the content will be properly cleaned up (just the boundaries will remain). What am I missing?

cibernox:
You didn't miss anything. I just wasn't sure if generating different html, even if it should be irrelevant could be considered breaking. 

I do not know why tests are not running tho.

rwjblue:
There was a fairly severe TravisCI outage (still recovering AFAICT). I tried to restart the build...

cibernox:
Ready to merge on my end. We could release as a minor version instead of a patch to be safe of any unintended consecuences?

rwjblue:
I'll try to get this published in the AM, unless @kratiahuja or others get to it first...

PR #188, opened 2018-06-07
v1.2.1 package upgrades

izelnakri:
tests passing


test/fastboot-headers-test.js
```diff
@@ -3,7 +3,7 @@
 var expect = require('chai').expect;
 var path = require('path');
 var FastBootHeaders = require('./../src/fastboot-headers.js');
-var Ember = require('ember-source/dist');
+var Ember = require('ember-source/dist/ember.debug');
```
kratiahuja: Not sure if we should be using debug version in production?

izelnakri: this is a test file and I dont see an index.js under `ember-source/dist`, I did this to make it more explicit what we are importing during the test.

rwjblue:
Rebased and fixed the conflicts, thanks again @izelnakri!

PR #189, opened 2018-06-16
Normalize headers to lower case

ryanto:
When accessing a header, fastboot [will lowercase the string](https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-headers.js#L51-L57) before looking up the header. However, when creating a `FastBootHeaders` object, the casing of the headers [is not changed](https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-headers.js#L20). This causes a `get` to a capitalized header to fail.

The docs use an example with a capitalized header, `headers.get('X-Request')`, but this will not work due to the case mismatch.

This PR normalizes all headers as lower case strings when creating a `FastBootHeaders` instance.  Now `X-Headers` will be accessible with `get`/`getAll` on `fastboot.request.headers`.

### DIFF
--- a/src/fastboot-headers.js
+++ b/src/fastboot-headers.js
@@ -17,7 +17,7 @@ class FastBootHeaders {
         value = [value];
       }
 
-      this.headers[header] = value;
+      this.headers[header.toLowerCase()] = value;
     }
   }
 
--- a/test/fastboot-headers-test.js
+++ b/test/fastboot-headers-test.js
@@ -6,6 +6,15 @@ var FastBootHeaders = require('./../src/fastboot-headers.js');
 var Ember = require('ember-source/dist');
 
 describe('FastBootHeaders', function() {
+  it('lower normalizes the headers to lowercase', function() {
+    var headers = {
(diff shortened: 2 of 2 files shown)


ryanto:
Not sure why travis is failing... looks like ubuntu repos are down. Can someone re-run?

kratiahuja:
Sorry for the delay! Re-running the job. I'll merge and release if the job passes.

ryanto:
Thanks @kratiahuja!

kratiahuja:
@ryanto One thing that slipped out of my mind: Given that we are storing the headers in lower case now, [we don't need to do lowercase during `getAll`](https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-headers.js#L51-L57). Can we remove it?

danmcclain:
As it’s user input, it would be safer to leave the toLowerCase call in GetAll, otherwise users always have to provide a lowercase string

kratiahuja:
Makes sense! Merged. Will release this week.

PR #190, opened 2018-06-24
Fix require path for on-disk modules

mhluska:
It looks like this path is incorrect (node_modules is not under dist but rather ../dist). It seems to be tripping up the ember-fetch library, see here: https://github.com/ember-cli/ember-fetch/issues/115


mhluska:
Closing since I don't think this works when deployed to Heroku.

PR #208, opened 2018-12-17
Fix to run all tests

kratiahuja:
Currently only one test (fastboot-dependencies) is running on master. This PR fixes to run all tests. Will be adding revelant eslint plugins as next PR so this doesn't happen in future.

cc: @rwjblue


Turbo87:
good catch! 👍 

we should probably use something like https://github.com/lo1tuma/eslint-plugin-mocha/ to prevent this from happening in the future

kratiahuja:
Yup I am exactly working on that next.

PR #225, opened 2019-07-12
Update node engines to match node versions we are testing against

kiwiupover:
We are testing against node 8, 10, 12 and stable
https://github.com/ember-fastboot/fastboot/blob/master/.travis.yml#L3-L6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
   - "stable"
-  - "6"
-  - "8"
+  - "12"
   - "10"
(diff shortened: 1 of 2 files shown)


rwjblue:
@kiwiupover The .travis.yml link you pasted above does include Node 6 (and that is what is failing in CI on this PR). I think you'll need to remove it.

kiwiupover:
@rwjblued does the node version `- "stable"` test against Node 13 when it is released of only Node 14, 16 and so on

PR #236, opened 2019-10-26
Refactor to use a single sandboxed context per visit request.

rwjblue:
* Make a `vm.createContext` per visit
* Remove the ability to provide a custom sandbox class (we can always
  add this back if needed, but the complexity it adds seems quite high
  and very very few (if any) folks use it).
* Ensure the scripts bytecode is shared between visit requests.
* Various internal refactors to simplify things as we went.

Co-authored-by: @krisselden

---

TODO:

- [x] Fix destroy process (identified during https://github.com/ember-fastboot/fastboot/pull/236#pullrequestreview-307546078)
- [x] Update `src/ember-app.js` to do a "simple" `app.boot()` to ensure that parsing failures themselves will fail the node process
- [x] Add tests showing prototype mutation does not "leak" per-visit (https://github.com/ember-fastboot/fastboot/pull/241)
- [x] Add tests showing `Ember.onerror` is unique per-visit (https://github.com/ember-fastboot/fastboot/pull/242)

Follow-up TODO:

* Migrate from `sandboxGlobals` to `buildSandboxGlobals` (https://github.com/ember-fastboot/fastboot/issues/239)
* Refactor `destroyAppInstanceInMs` (https://github.com/ember-fastboot/fastboot/issues/240)

---

Closes #241 
Closes #242


src/ember-app.js
```diff
@@ -365,22 +319,27 @@ class EmberApp {
...
+      // eslint-disable-next-line require-atomic-updates
+      result.error = error;
+    } finally {
+      if (result._destroyAppInstance()) {
+        if (destroyAppInstanceTimer) {
+          clearTimeout(destroyAppInstanceTimer);
         }
-      });
+      }
+    }
```
kratiahuja: Given we are creating a new `EmberApp` every `visit` now, this means that we can only have 1 App Instance per `EmberApp` rather previously n EmberAppInstances per EmberApp. Please correct me if I am misunderstanding this. 

Given the above, we should also destroy the `EmberApp` after every `visit` (by calling and fixing the `destroy` method) so that any timer related code doesn't run until the next `visit` request.  Otherwise we will end up running async code without a live app instance and not be able to attribute those async errors that occur after visit correctly.

src/ember-app.js
```diff
@@ -234,30 +203,13 @@ class EmberApp {
...
-    return this.createEmberApp();
-  }
-
   /**
    * Destroys the app and its sandbox.
    */
   destroy() {
     if (this.app) {
       this.app.destroy();
     }
```
kratiahuja: I think `this.app` doesn't exists in the new changes any more so this condition will always be `false`.

src/ember-app.js
```diff
@@ -267,11 +219,22 @@ class EmberApp {
...
+    debug('adding files to sandbox');
+
+    for (let script of this.scripts) {
+      debug('evaluating file %s', script);
+      sandbox.runScript(script);
+    }
+
+    debug('files evaluated');
+
+    let app = await this.createEmberApp(sandbox).boot();
```
kratiahuja: I think you may want to `this.app` instead of a local variable here.

src/ember-app.js
rwjblue: > Given the above, we should also destroy the EmberApp and the sandbox after every visit (by calling and fixing the destroy method) so that any timer related code doesn't run until the next visit request.

Yep, totally correct. Will update to ensure we call destroy.

src/ember-app.js
rwjblue: I was trying to avoid _ever_ setting `this.app`, specifically to ensure that we don't "leak" it or accidentally allow it to live longer than it should.

I'll definitely fix up the destroy code though (which you called out in another comment above).

src/ember-app.js
rwjblue: > Given we are creating a new EmberApp every visit now, this means that we can only have 1 App Instance per EmberApp at any given point of time rather previously n EmberAppInstances per EmberApp. 

Actually this isn't _quite_ correct, but the terminology actually confused me. 😄 I'm going to refer to "ember app" as the actual app that is booted (from `distPath`) and "fastboot ember app" as this local file (`src/ember-app.js`). 

The "fastboot ember app" is instantiated once (by the main entry point in `src/index.js`) when the end user calls `let fastboot = new FastBoot({ distPath })` and is destroyed then re-created whenever the user calls `fastboot.reload()`. The "ember app" will be instantiated once per `visit` call, and destroyed upon completion.

rwjblue: OK, just pushed a commit that ensures both `Ember.Application` and `Ember.ApplicationInstance` are destroyed after the visit concludes.

rwjblue: OK, just pushed a commit that ensures both Ember.Application and Ember.ApplicationInstance are destroyed after the visit concludes.

src/ember-app.js
```diff
@@ -180,43 +185,28 @@ class EmberApp {
...
-    debug('vendor file evaluated');
-
-    appFilePaths.forEach(function(appFilePath) {
-      debug('evaluating app file %s', appFilePath);
-      let appFile = fs.readFileSync(appFilePath, 'utf8');
-      sandbox.eval(appFile, appFilePath);
-    });
-    debug('app files evaluated');
-  }
+  destroy() {}
```
kratiahuja: I think the `reload` function still calls this method: https://github.com/ember-fastboot/fastboot/blob/master/src/index.js#L93, should we just remove it?

kratiahuja:
@rwjblue Thanks for making these changes, these look good to me except the two questions above.

src/ember-app.js
rwjblue: I don't want to remove it yet, because I want to make it do something in a follow up PR (but before v3.0.0). Specifically, I want to "warm" the cache even more which means I'll have to be able to cleanup some of the async.

src/result.js
```diff
@@ -150,15 +150,20 @@ class Result {
...
-    return false;
+    this.isDestroyed = true;
+
+    if (this.applicationInstanceInstance !== undefined) {
+      this.applicationInstanceInstance.destroy();
+    }
+
+    if (this.applicationInstance !== undefined) {
+      this.applicationInstance.destroy();
+    }
```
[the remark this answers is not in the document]
rwjblue: @kratiahuja - This removes the conditionalness (if that is a word), we _always_ call destroy and what we used to use the return value for isn't needed any more. Am I missing something?

src/ember-app.js
kratiahuja: aah ok can we add a todo then here?

src/result.js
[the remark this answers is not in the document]
kratiahuja: I think the place where we destroy the application with a timer (`destroyAppInstanceInMs`): `if (result._destroy()) {`, that would never run because we aren't return a `true` value.

src/ember-app.js
```diff
@@ -345,42 +299,44 @@ class EmberApp {
...
-    });
-
+    // TODO: Use Promise.race here
     let destroyAppInstanceTimer;
     if (destroyAppInstanceInMs > 0) {
       // start a timer to destroy the appInstance forcefully in the given ms.
       // This is a failure mechanism so that node process doesn't get wedged if the `visit` never completes.
       destroyAppInstanceTimer = setTimeout(function() {
-        if (result._destroyAppInstance()) {
+        if (result._destroy()) {
```
kratiahuja: This condition will always be false for visits that never completed so we will never set `result.error`

src/result.js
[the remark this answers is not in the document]
kratiahuja: Put a new comment to point to the actual code which will never run.

src/ember-app.js
rwjblue: Aha! thank you, will fix this (I don't think it will need to be conditional in the end once we race, but in the meantime I'll bring back the true/false).

kratiahuja: yeah that is why I asked :)

src/ember-app.js
rwjblue: Yep! Will do.

src/ember-app.js
rwjblue: OK, fixed this with the latest commit.

PR #245, opened 2019-10-31
Refactor sandboxGlobals -> buildSandboxGlobals

### LINKED ISSUE #239: Refactor `sandboxGlobals` options into `buildSandboxGlobals`
Now that we are not sharing a single sandbox context for all visits, we will need to change the API to provide custom globals from a single stable object with properties to be assigned into the VM sandbox to a builder function that will return a new object for each request.

Specifically, we will move from:

```js
let app = new FastBoot({
  distPath: 'path/to/dist',
  sandboxGlobals: { someProperty: whateverHere },
});
```

To:

```
let app = new FastBoot({
  distPath: 'path/to/dist',
  buildSandboxGlobals() { 
    return { someProperty: whateverHere };
  },
});
```

rwjblue:
This changes the system from providing default set of shared (and therefore mutable) global properties to using a builder function to generate the set of globals to be used _per visit_.

The `buildSandboxGlobals` function will receive the default set of globals that FastBoot creates (currently this is `najax` and `FastBoot`), and whatever the `buildSandboxGlobals` function returns is what will ultimately be used. If `buildSandboxGlobals` is not provided, a default implementation is used (it is essentially `defaultGlobals => defaultGlobals;`).

Closes #239

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,14 @@ let app = new FastBoot({
   distPath: 'path/to/dist',
   // optional boolean flag when set to true does not reject the promise if there are rendering errors (defaults to false)
   resilient: <boolean>,
-  sandboxGlobals: {...} // optional map of key value pairs to expose in the sandbox
+
+  // optional function used to generate the set of global properties available within the sandbox, receives default globals
+  // and is expected to return an object (the default implementation returns the passed in defaults)
+  buildSandboxGlobals(defaultGlobals) {
+    return Object.assign({}, defaultGlobals, {
+      // additional global properties to define within the sandbox
+    });
+  },
 });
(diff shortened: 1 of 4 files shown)


krisselden:
We could default it to a function that does a cheap copy of the old property to ease migration.

Making it a function just helps clue people into that they should be careful here, it doesn't mitigate data leaking between visits and the existing code already assigns.  This doesn't prevent interior mutations and we already Object.assign the old sandboxGlobals property.

rwjblue:
Updated to default `buildSandboxGlobals` from a specified `options.sandboxGlobals` (with a console warning).

PR #247, opened 2019-10-31
Remove najax from default set of sandbox globals.

### LINKED ISSUE #246: Drop `najax` global by default?
As we continue work towards the next major version (will be v3.0.0), I'd like to seriously consider dropping `najax` from the default set of global properties added by this library.

The primary reasons are:

* Recent versions of ember-data (3.12+) will automatically use `ember-fetch` (or global `fetch`) if present, and does not need special logic around `najax`
* New applications (as of ember-cli@3.13.0) include `ember-fetch` by default, and do **not** include jQuery by default. Considering that `najax` is a `jQuery.ajax` emulation API, I think this library should avoid exposing it.
* `najax` (as a dependency of this library) is difficult for the host application to control (e.g. get their own version), and would be better if they provided it themselves (via `buildSandboxGlobals` API)
* Providing backwards compatibility is **very** easy (add `najax: require('najax')` via `buildSandboxGlobals`)
* It seems better to "follow a spec" (suggesting `fetch` usage)

Thoughts?

rwjblue:
For applications that still need `najax` (for example, if they are using
an older ember-data version), they can do (example uses new
`buildSandboxGlobals` API):

```js
const najax = require('najax');

let fastboot = new FastBoot({
  distPath: 'some/path/to/dist',
  buildSandboxGlobals(globals) {
    return Object.assign({}, globals, {
      najax,
    });
  }
});
```

See #246 for more details about this change.

Closes #246

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "chalk": "^2.4.2",
     "cookie": "^0.4.0",
     "debug": "^4.1.1",
-    "najax": "^1.0.4",
     "resolve": "^1.12.0",
     "simple-dom": "^1.4.0",
     "source-map-support": "^0.5.13"
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -5,7 +5,6 @@ const vm = require('vm');
 const path = require('path');
 const chalk = require('chalk');
 
-const najax = require('najax');
 const SimpleDOM = require('simple-dom');
 const resolve = require('resolve');
 const debug = require('debug')('fastboot:ember-app');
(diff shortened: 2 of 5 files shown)


runspired:
fwiw you need najax even in current versions of ember-data if you don't have ember-fetch installed. This change trolled our fastboot tests in ember-data itself because when we test the jquery scenario we add @ember/jquery and remove ember-fetch, once we bump to fastboot 3 this fails (see https://github.com/embermap/ember-cli-fastboot-testing/issues/471 and https://github.com/ember-fastboot/ember-cli-fastboot/issues/829).

It's not clear how to add najax back in from this example for ember-cli-fastboot-testing.

PR #252, opened 2019-11-04
Expose option to allow a new sandbox per visit

rwjblue:
This PR does a few things:

Firstly, it allows the end consumer to decide if they want a sandbox per visit or to share a single sandbox over all visits.

Second, it changes the default of `buildSandboxPerVisit` back to `false` therefore sharing a single sandbox for many `visit` invocations.

Reverts the defaults change introduced in #236.

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -183,8 +183,9 @@ class EmberApp {
    * Perform any cleanup that is needed
    */
   destroy() {
-    // TODO: expose as public api (through the top level) so that we can
-    // cleanup pre-warmed visits
+    if (this._applicationInstance) {
+      this._applicationInstance.destroy();
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -75,6 +75,7 @@ class FastBoot {
    * @param {Boolean} [options.shouldRender] whether the app should do rendering or not. If set to false, it puts the app in routing-only.
    * @param {Boolean} [options.disableShoebox] whether we should send the API data in the shoebox. If set to false, it will not send the API data used for rendering the app on server side in the index.html.
    * @param {Integer} [options.destroyAppInstanceInMs] whether to destroy the instance in the given number of ms. This is a failure mechanism to not wedge the Node process (See: https://github.com/ember-fastboot/fastboot/issues/90)
+   * @param {Boolean} [options.buildSandboxPerVisit] whether to create a new sandbox context per-visit (slows down each visit, but guarantees no prototype leakages can occur), or reuse the existing sandbox (faster per-request, but each request shares the same set of prototypes)
```
kratiahuja: Can we put a default value here?

src/ember-app.js
```diff
@@ -242,12 +243,27 @@ class EmberApp {
...
+    if (buildSandboxPerVisit) {
+      // entangle the specific application instance to the result, so it can be
+      // destroyed when result._destroy() is called (after the visit is
+      // completed)
+      result.applicationInstance = app;
+    } else {
+      // save the created application instance so that we can clean it up when
+      // this instance of `src/ember-app.js` is destroyed (e.g. reload)
+      this._applicationInstance = app;
+    }
```
kratiahuja: I think this is a bit confusing in terms of variable naming. IIUC, we are deciding here whether to build a new EmberApp every request or re-use the old one and still create a new EmberAppInstance. Please correct me if I am wrong.

If that is the case, shouldn't we call it `this._application` and `result.application`. It isn't an instance per say. I know the already existing code calls it `appInstance` and appInstanceInstance` but that can easily cause confusion :)

rwjblue: > I think this is a bit confusing in terms of variable naming.

Yeah, I don't disagree, but we are talking about instances of `Ember.Application` **and** `Ember.ApplicationInstance`. We have to call them _something_...

>  It isn't an instance per say.

Well, it is actually. It is the result of `Ember.Application.create()`, it is an `Ember.Application` instance (hence the name `this._applicationInstance`). The other thing we get is an `Ember.ApplicationInstance` instance (and is called `this._applicationInstanceInstance`).

---

tldr; I erred on the side of "correct but overly verbose", I could change it to `.application` and `.instance` if you prefer.

kratiahuja: yeah my only concern was around accidentally misread of the variables due to repeated subwords.

rwjblue: Yeah :( including `Instance` in the name of a class is bad news bears 😭

dnalagatla: why not use `app` as variable name?  I agree using `appInstance` as a variable name is a bit confusing while reading the code.

rwjblue: “App” is significantly overused as well. Is it an instance of `src/ember-app.js` in this project? Is it an instance of `lib/broccoli/ember-app.js` in ember-cli? Is it an Ember.Application instance? Is it the conceptual “application” that we are rendering? 🙀 

Tbh, I’d rather make up completely fabricated names (that don’t use “instance” at all) and link to a naming doc to explain what each of the names mean.

dnalagatla: I agree, it's a tough problem.

PR #262, opened 2020-03-01
Build new sandbox after request is done

kratiahuja:
This is basic PR to build a new sandbox per visit after current request is complete. Specifically making sure the creation of sandbox is not tied to the request timeline itself but is done as an after effect. This is so that we do not build sandbox at the start of the request which is slow.

Not specifically adding tests as this is just an internal implementation change. When we add queue management, I'll add more test cases of requests more than queue size or less than to cover those usecases.

cc: @rwjblue

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -189,6 +189,16 @@ class EmberApp {
   }
 
   /**
+   * Builds a new application instance sandbox as a micro-task.
+   */
+  buildNewApplicationInstance() {
+    return Promise.resolve().then(() => {
+      let app = this.buildApp();
+      return app;
+    });
+  }
+
+  /**
    * @private
    *
    * Creates a new `Application`
@@ -230,6 +240,29 @@ class EmberApp {
   /**
    * @private
    *
+   * Get the new sandbox off if it is being created, otherwise create a new one on demand.
+   * The later is needed when the current request hasn't finished or wasn't build with sandbox
+   * per request turned on and a new request comes in.
+   *
+   */
+  async _getNewApplicationInstance() {
+    let app;
+
+    if (this._pendingNewApplicationInstance) {
+      let pendingAppInstancePromise = this._pendingNewApplicationInstance;
+      this._pendingNewApplicationInstance = undefined;
+      app = await pendingAppInstancePromise;
+    } else {
+      // if there is no current pending application instance, create a new one on-demand.
+      app = await this.buildApp();
+    }
+
+    return app;
+  }
+
+  /**
+   * @private
+   *
    * Main function that creates the app instance for every `visit` request, boots
    * the app instance and then visits the given route and destroys the app instance
    * when the route is finished its render cycle.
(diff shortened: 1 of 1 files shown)


src/ember-app.js
```diff
@@ -227,6 +237,28 @@ class EmberApp {
...
+   * The later is needed when the current request hasn't finished or wasn't build with sandbox
+   * per request turned on and a new request comes in.
+   *
+   */
+  async _getNewApplicationInstance() {
+    let app;
+
+    if (this._pendingNewApplicationInstance) {
+      app = await this._pendingNewApplicationInstance;
+      this._pendingNewApplicationInstance = undefined;
```
rwjblue: I think we need to reset this to undefined _before_ waiting (otherwise we have a race condition and sending many pending requests at the same time might force the same sandbox to be used accidentally).

kratiahuja:
I also quickly verified the time for `visit` and creation of sandbox has intended effect:

```
  FastBoot
DEBUG: -------------------------------
DEBUG: Ember : 3.14.1
DEBUG: -------------------------------
DEBUG: -------------------------------
DEBUG: Ember : 3.14.1
DEBUG: -------------------------------
TIME_TO_CREATE_SANDBOX: 13.023ms
TIME_FOR_VISIT: 46.644ms
DEBUG: -------------------------------
DEBUG: Ember : 3.14.1
DEBUG: -------------------------------
TIME_TO_CREATE_SANDBOX: 0.012ms
TIME_FOR_VISIT: 9.980ms
DEBUG: -------------------------------
DEBUG: Ember : 3.14.1
DEBUG: -------------------------------
TIME_TO_CREATE_SANDBOX: 0.033ms
TIME_FOR_VISIT: 9.907ms
DEBUG: -------------------------------
DEBUG: Ember : 3.14.1
DEBUG: -------------------------------
    ✓ in app prototype mutations do not leak across visits with buildSandboxPerVisit=true (260ms)
```

src/ember-app.js
```diff
@@ -355,6 +383,12 @@ class EmberApp {
       result._destroy();
 
       clearTimeout(destroyAppInstanceTimer);
+
+      if (buildSandboxPerVisit) {
+        // if sandbox was built for this visit, then build a new sandbox for the next incoming request
+        // which is invoked using buildSandboxPerVisit
+        this._pendingNewApplicationInstance = this.buildNewApplicationInstance();
```
melmerp: I might be reading this wrong, but if there are multiple concurrent `visit()` calls then it's possible to call this twice. I think you might need to make sure that `this._pendingNewApplicationInstance` is empty before doing this. If it is already set, then you probably shouldn't do anything.

This leads me to believe that there will be at most one application instance waiting to be used, so if multiple visit calls come at the same time, the second one will need to build the app instance anyways.

kratiahuja: Thanks for catching that. I'll fix this in a subsequent PR.

src/ember-app.js
```diff
@@ -188,6 +188,16 @@ class EmberApp {
     }
   }
 
+  /**
+   * Builds a new application instance sandbox as a micro-task.
+   */
+  buildNewApplicationInstance() {
```
melmerp: Should this function be marked as '@private'? What's the convention for using '_' for naming a method?

PR #264, opened 2020-03-08
Add sandbox queue management when using buildSandboxPerVisit

kratiahuja:
In order to increase the efficiency of the system when using `buildSandboxPerVisit` flag, we currently lazily create the sandbox for the next request. This is a single sandbox that is leveraged. In order to help with performance, we will be maintaining a queue of these sandboxes that can be leveraged by the requests. When the sandbox is empty, it will eagerly create the sandbox instead of caching the pending request and letting it run when the sandbox is available. This is primarily so that we don't end up starving the system when number of requests is far greater than queue size and the system cannot recover from storing the sandbox for next requests. Finishing the request (even if in degraded fashion) brings in better resilience.

In addition, when requests cannot use the sandbox from the queue, we now emit analytics information about the request in the result object. This helps applications using this flag to track if their QPS is greater than Queue size and bump the queue size.

The queue size is provided when the app is created and pre-emptively the queue is filled up so initial requests don't starve.

Updated README as well.

cc: @rwjblue @melmerp

Next: 
- [ ] Add visit timing information in analytics of the result object


kratiahuja:
Also ran some perf numbers against the tests and they are in-line to the previous PR [here](https://github.com/ember-fastboot/fastboot/pull/262) when sandbox is being leveraged from a previous request.

src/utils/queue.js
```diff
@@ -0,0 +1,40 @@
...
+    return this.items.length === 0;
+  }
+
+  size() {
+    return this.items.length;
+  }
+
+  isFull() {
+    return this.items.length === this.maxSize;
+  }
```
rileyhilliard: Opportunity to leverage the Queues `this.size()` method for all the `this.items.length` checks. It looks like this method is only used for testing, but leveraging it in the actual code makes things a tad more DRY.

src/ember-app.js
```diff
@@ -66,7 +68,23 @@ class EmberApp {
...
     // Ensure that the dist files can be evaluated and the `Ember.Application`
     // instance created.
-    this.buildApp();
+    this.buildSandboxQueue(options.maxSandboxQueue);
+  }
+
+  /**
+   * @private
+   *
+   * unction to build queue of sandboxes which is later leveraged if application is using buildSandboxPerRequest
```
rileyhilliard: (spelling)
_function to build queue of sandboxes which is later leveraged if the application is using buildSandboxPerRequest_

src/utils/queue.js
kratiahuja: Good point, let me update this. Thanks for catching this.

src/ember-app.js
kratiahuja: Thanks for catching this.

src/ember-app.js
```diff
@@ -285,13 +303,19 @@ class EmberApp {
...
+
+    let app = shouldBuildApp ? await this.getNewApplicationInstance() : this._applicationInstance;
 
     if (buildSandboxPerVisit) {
       // entangle the specific application instance to the result, so it can be
       // destroyed when result._destroy() is called (after the visit is
       // completed)
       result.applicationInstance = app;
+
+      // we add analytics information about the current request to know where it is degraded or not.
```
melmerp: "where it is" -> "whether it is"

src/utils/queue.js
```diff
@@ -0,0 +1,45 @@
...
+  constructor(maxSize = 1) {
+    this.items = [];
+    this.maxSize = maxSize;
+  }
+
+  enqueue(item) {
+    if (this.size() < this.maxSize) {
+      this.items.push(item);
+    } else {
+      debug('Ignoring adding appInstance to queue as Queue is already full!');
```
melmerp: Why silently ignore? Shouldn't this throw an error or return some value indicating that you're not getting what you want?

Or you should document that you must check `isFull()` before calling this to ensure that there's space. Otherwise, I don't know what to expect.

src/utils/queue.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  dequeue() {
+    if (!this.isEmpty()) {
+      return this.items.shift();
+    }
+
+    return null;
```
melmerp: Is `null` allowed to be an item in the queue, if so, then maybe document that `null` means either the item is null or the queue is empty. Or you can just disallow nulls somehow.

README.md
```diff
@@ -36,6 +36,10 @@ let app = new FastBoot({
       // additional global properties to define within the sandbox
     });
   },
+
+  // optional number to be provided when using `buildSandboxPerVisit` which defines the queue size for sandboxes.
+  // This number should represent your QPS of your service
+  maxSandboxQueue: <Number> // defaults to 1 if not provided
```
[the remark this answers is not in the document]
rwjblue: Ya, agree.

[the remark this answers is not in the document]
rwjblue: Also, I don't quite think it defaults to 1 (as that would infer, at least to me, that we would process that one in a separate thread).

README.md
```diff
@@ -107,6 +111,14 @@ rendering your Ember.js application using the [FastBoot App Server](https://gith
 Run `fastboot` with the `DEBUG` environment variable set to `fastboot:*`
 for detailed logging.
 
+### Result
+
+The result from `fastboot` is a `Result` object that has the following API:
+
+1. `html`: this returns the serialized DOM contents after running the route on server side
+2. `domContents` : This returns an object that contains the `head` and `body` of the server side rendered content in a serialized string
+3. `getAnalytics`: This returns any request specfic analytical information that can be leveraged for monitoring the application. Example: when using `buildSandboxPerVisit`, this returns if the request ran in a degraded fashion when the queue was starved.
```
rwjblue: Thoughts on tweaking this to be in a TypeScript interface presentation?

```ts
interface FastBootVisitResult {
  /**
    The serialized DOM contents after completing the `visit` request.

    Note: this combines the `domContents.head` and `domContents.body`.
  */
  html: string;
  
  domConents: {
    /**
      The `<head>` contents generated by the visit.
    */
    head: string;

    /**
      The `<body>` contents generated by the visit.
    */
    body: string;
  };
}
```

I do not fully understand the `getAnalytics` part of the result. Is it a function the user has to call? Is it an object?

src/ember-app.js
```diff
@@ -66,7 +68,23 @@ class EmberApp {
...
+    this.buildSandboxQueue(options.maxSandboxQueue);
+  }
+
+  /**
+   * @private
+   *
+   * Function to build queue of sandboxes which is later leveraged if application is using `buildSandboxPerRequest`
+   * flag. This is an optimization to help with performance.
+   *
+   * @param {Number} maxSandboxQueueSize - maximum size of queue (this is should be a derivative of your QPS)
```
rwjblue: > maxSandboxQueueSize

Can we we use this name throughout?

src/ember-app.js
```diff
@@ -285,13 +303,19 @@ class EmberApp {
...
+    let app = shouldBuildApp ? await this.getNewApplicationInstance() : this._applicationInstance;
 
     if (buildSandboxPerVisit) {
       // entangle the specific application instance to the result, so it can be
       // destroyed when result._destroy() is called (after the visit is
       // completed)
       result.applicationInstance = app;
+
+      // we add analytics information about the current request to know where it is degraded or not.
+      result.analytics.sandboxPerVisit.isDegradedRequest = isSandboxEagerlyBuilt;
```
rwjblue: I don't think "degraded" has the right meaning here. There are _many_ reasons a given request may have a degraded processing time, this is only representing one of them.

Some possible alternate names:

* `result.analytics.sandboxPerVisit.handledByQueue`
* `result.analytics.sandboxPerVisit.sandboxPrebuilt`
* ???

src/ember-app.js
```diff
@@ -384,10 +408,10 @@ class EmberApp {
 
       clearTimeout(destroyAppInstanceTimer);
 
-      if (buildSandboxPerVisit) {
+      if (buildSandboxPerVisit && !this._sandboxApplicationInstanceQueue.isFull()) {
         // if sandbox was built for this visit, then build a new sandbox for the next incoming request
         // which is invoked using buildSandboxPerVisit
-        this._pendingNewApplicationInstance = this.buildNewApplicationInstance();
+        this._sandboxApplicationInstanceQueue.enqueue(this.buildNewApplicationInstance());
```
rwjblue: I'd prefer the queue to manage this state internally (avoiding the `isFull` check here).

src/ember-app.js
```diff
@@ -66,7 +68,23 @@ class EmberApp {
...
+  /**
+   * @private
+   *
+   * Function to build queue of sandboxes which is later leveraged if application is using `buildSandboxPerRequest`
+   * flag. This is an optimization to help with performance.
+   *
+   * @param {Number} maxSandboxQueueSize - maximum size of queue (this is should be a derivative of your QPS)
+   */
+  buildSandboxQueue(maxSandboxQueueSize = 1) {
+    this._sandboxApplicationInstanceQueue = new Queue(maxSandboxQueueSize);
```
rwjblue: Can we change the Queue constructor to something like:

```js
new Queue(() => this.buildNewApplicationInstance(), totalSize)
```

Then the queue would "fill" itself.

src/ember-app.js
```diff
@@ -245,16 +263,16 @@ class EmberApp {
...
+  async getNewApplicationInstance() {
     let app;
 
-    if (this._pendingNewApplicationInstance) {
-      let pendingAppInstancePromise = this._pendingNewApplicationInstance;
-      this._pendingNewApplicationInstance = undefined;
-      app = await pendingAppInstancePromise;
-    } else {
-      // if there is no current pending application instance, create a new one on-demand.
+    if (this._sandboxApplicationInstanceQueue.isEmpty()) {
```
rwjblue: I'd prefer to always get the instance from the queue, and the queue would return a structured object that indicates if it had to "build on demand" or had a pre-warmed instance (related to comment about RE: passing the "builder" function into the `Queue` constructor).

src/result.js
```diff
@@ -111,6 +114,15 @@ class Result {
     };
   }
 
+  /**
+   * Get request specific analytics information.
+   *
+   * @returns {Object} analytics about the current request
+   */
+  getAnalytics() {
```
rwjblue: I do not understand the point of this function, why do we need a `result.getFoo()` when we could use `result.foo` instead?

test/fastboot-test.js
```diff
@@ -409,19 +409,25 @@ describe('FastBoot', function() {
     });
 
     let result = await fastboot.visit('/', { buildSandboxPerVisit: true });
+    let analytics = result.getAnalytics();
     let html = await result.html();
 
     expect(html).to.match(/Items: 1/);
+    expect(analytics.sandboxPerVisit.isDegradedRequest).to.be.false;
```
rwjblue: Can we change these tests to deep equals (asserting the full contents of `analytics`)?

README.md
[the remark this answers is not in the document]
kratiahuja: I'll update this to `maxSandboxQueueSize`. Not sure I understand why it doesn't default to 1? I think eventually also we need one web worker thread regardless IMO. Please let me knwo if I am misunderstanding something.

src/utils/queue.js
kratiahuja: The reason to silently ignore is and not throw an error is because when the number of requests is more than the size of the queue, we anyways process the request by creating a sandbox on demand. Throwing an error will block us from doing that. 

Sure I'll document calling `isFull()`.

src/utils/queue.js
kratiahuja: Sure I'll document `null` and what it means here.

README.md
kratiahuja: That is a good suggestion. I'll update it. 

Yes `getAnalytics` is a public function that the user calls to get any request specific info that was built during `visit`. The return value is an object. Alternatively, you could just call `result.analytics` as well, this just provides a getter function.

In future, I want to add more timing info on time took for `visit` and other measurements that apps can track.

src/result.js
kratiahuja: We can do either, I was more inclined to putting a getter function but can do `result.foo` as well.

src/ember-app.js
kratiahuja: I am not sure I can do that unless I leak some details into the queue. The instance differs when it is returned from the queue and when it is built on demand. The instance in a queue is a promise which resolves to the sandbox while build on demand is an instance directly.

basically when it is pre-warmed we call `Promise.resolve()...` while on demand it is `buildApp()` which are two different functions.

src/ember-app.js
kratiahuja: Thanks for the feedback. Agree with you. I ended up with `usedPrebuiltSandbox`

Issue #107, opened 2017-01-06
FastBoot should emit metrics


kratiahuja:
This issue is an *enhancement* for future purposes.

Post FastBoot 1.0, we should look at emmitting some useful metrics like server side latency, memory used by the node processes so that apps using FastBoot can listen to these metrics and create useful operational dashboard. 

This probably will need an RFC at some point IMO.

sdhull:
notes from discussion in fastboot meeting:
* would tooling be low-overhead and on all the time?
* would there be some debugging mode that you can turn on just to debug?
* are there specific metrics you want?
 * time from when the visit request started to when the promise was resolved
 * when rehydration starts, how long does it take?
* we're not going to do better than existing node.js tools re: memory leaks & stuff
* express & node have tooling
* what kind of unique-to-ember things can we emit?
 * did a rerender get scheduled?
 * did you use a tag that is known to have issues with rehydration? (eg `<select>` [aside: wait, for real?])

@ryanto has some middlewares for some tooling he might be willing to share...?

rondale-sc:
I believe in the rehydration builder there is a path that happens when a node is skipped or removed (when the rehydration builder can't decipher something so it throws it out).  It would be great to see a counter of that accessible in some sort of debug mode to determine how effective rehydration is.

PR #276, opened 2020-10-07
Add instrumentation for measuring metrics related to visit

suchitadoshi1987:
Added instrumentation to measure
- time taken to complete the fastboot visit
- time taken to complete the instance visit
- time taken to retrieve the app instance from the queue

### DIFF
--- a/src/ember-app.js
+++ b/src/ember-app.js
@@ -188,8 +188,11 @@ class EmberApp {
    *
    * @returns {Object}
    */
-  getAppInstanceInfo(appInstance, isAppInstancePreBuilt = true) {
-    return { app: appInstance, isSandboxPreBuilt: isAppInstancePreBuilt };
+  getAppInstanceInfo(
+    appInstance,
+    appAnalytics = { isAppInstancePreBuilt: true, appInstanceRetrievalTime: 0 }
+  ) {
+    return { app: appInstance, appAnalytics };
   }
 
   /**
@@ -205,10 +208,14 @@ class EmberApp {
    *                                       only
    */
   async getNewApplicationInstance() {
+    const appInstanceRetrievalStartTime = Date.now();
     const queueObject = this._sandboxApplicationInstanceQueue.dequeue();
     const app = await queueObject.item;
 
-    return this.getAppInstanceInfo(app, queueObject.isItemPreBuilt);
+    return this.getAppInstanceInfo(app, {
+      isAppInstancePreBuilt: queueObject.isItemPreBuilt,
+      appInstanceRetrievalTime: Date.now() - appInstanceRetrievalStartTime,
+    });
   }
 
   /**
@@ -236,10 +243,11 @@ class EmberApp {
   async _visit(path, fastbootInfo, bootOptions, result, buildSandboxPerVisit) {
     let shouldBuildApp = buildSandboxPerVisit || this._applicationInstance === undefined;
 
-    let { app, isSandboxPreBuilt } = shouldBuildApp
+    let { app, appAnalytics } = shouldBuildApp
       ? await this.getNewApplicationInstance()
       : this.getAppInstanceInfo(this._applicationInstance);
 
+    result.analytics.appInstanceRetrievalTime = appAnalytics.appInstanceRetrievalTime;
     if (buildSandboxPerVisit) {
       // entangle the specific application instance to the result, so it can be
       // destroyed when result._destroy() is called (after the visit is
@@ -248,7 +256,7 @@ class EmberApp {
 
       // we add analytics information about the current request to know
       // whether it used sandbox from the pre-built queue or built on demand.
-      result.analytics.usedPrebuiltSandbox = isSandboxPreBuilt;
+      result.analytics.usedPrebuiltSandbox = appAnalytics.isAppInstancePreBuilt;
     } else {
       // save the created application instance so that we can clean it up when
       // this instance of `src/ember-app.js` is destroyed (e.g. reload)
@@ -262,7 +270,11 @@ class EmberApp {
     registerFastBootInfo(fastbootInfo, instance);
 
     await instance.boot(bootOptions);
+
+    const instanceVisitStartTime = Date.now();
     await instance.visit(path, bootOptions);
+    result.analytics.instanceVisitTime = Date.now() - instanceVisitStartTime;
+
     await fastbootInfo.deferredPromise;
   }
 
@@ -290,6 +302,7 @@ class EmberApp {
    * @returns {Promise<Result>} result
    */
   async visit(path, options) {
+    let fastbootVisitStartTime = Date.now();
     let req = options.request;
     let res = options.response;
     let html = options.html || this.html;
@@ -323,6 +336,7 @@ class EmberApp {
 
     try {
       await this._visit(path, fastbootInfo, bootOptions, result, buildSandboxPerVisit);
+      result.analytics.fastbootVisitTime = Date.now() - fastbootVisitStartTime;
 
       if (!disableShoebox) {
         // if shoebox is not disabled, then create the shoebox and send API data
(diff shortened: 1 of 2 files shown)


src/ember-app.js
```diff
@@ -205,10 +208,14 @@ class EmberApp {
    *                                       only
    */
   async getNewApplicationInstance() {
+    const appInstanceRetrievalStartTime = Date.now();
```
rwjblue, line 211: I'd generally prefer to use something with higher precision. Mind swapping to `performance.now()` ([docs here](https://nodejs.org/dist/latest-v14.x/docs/api/perf_hooks.html#perf_hooks_performance_now)):

```js
const { performance } = require('perf_hooks');

// ...snip...
const start = performance.now();
```

src/ember-app.js
```diff
@@ -205,10 +208,14 @@ class EmberApp {
...
     const queueObject = this._sandboxApplicationInstanceQueue.dequeue();
     const app = await queueObject.item;
 
-    return this.getAppInstanceInfo(app, queueObject.isItemPreBuilt);
+    return this.getAppInstanceInfo(app, {
+      isAppInstancePreBuilt: queueObject.isItemPreBuilt,
+      appInstanceRetrievalTime: Date.now() - appInstanceRetrievalStartTime,
```
rwjblue, line 217: I'd like to store the start and stops too. Also can we rename this to `buildAppInstanceTiming`?

Something like:
```js
const start = performance.now();
const queueObject = this._sandboxApplicationInstanceQueue.dequeue();
const app = await queueObject.item;
const stop = performance.now();

return this.getAppInstanceInfo(app, {
  isAppInstancePreBuilt: queueObject.isItemPreBuilt,
  buildAppInstanceTiming: {
    start,
    stop,
    duration: stop - start,
  }
});
```

src/ember-app.js
```diff
@@ -236,10 +243,11 @@ class EmberApp {
   async _visit(path, fastbootInfo, bootOptions, result, buildSandboxPerVisit) {
     let shouldBuildApp = buildSandboxPerVisit || this._applicationInstance === undefined;
 
-    let { app, isSandboxPreBuilt } = shouldBuildApp
+    let { app, appAnalytics } = shouldBuildApp
```
rwjblue, line 246: Can we call this `analytics` instead of `appAnalytics`?

rwjblue: Also, can we unify these? Can we funnel both cases through a single function (e.g. `buildAppInstance` or even `getNewApplicationInstance`) so the rest of the code doesn't need to care about the two pathways?

src/ember-app.js
```diff
@@ -262,7 +270,11 @@ class EmberApp {
     registerFastBootInfo(fastbootInfo, instance);
 
     await instance.boot(bootOptions);
+
+    const instanceVisitStartTime = Date.now();
     await instance.visit(path, bootOptions);
+    result.analytics.instanceVisitTime = Date.now() - instanceVisitStartTime;
```
rwjblue, line 276: Similar to above, lets use a POJO to group multiple values:

```js
const start = performance.now();
await instance.visit(path, bootOptions);
const stop = performance.now();
result.analytics.instanceVisitTimings = {
  start,
  stop,
  duration: stop - start,
};
```

src/ember-app.js
```diff
@@ -323,6 +336,7 @@ class EmberApp {
 
     try {
       await this._visit(path, fastbootInfo, bootOptions, result, buildSandboxPerVisit);
+      result.analytics.fastbootVisitTime = Date.now() - fastbootVisitStartTime;
```
rwjblue, line 339: Same as above

rwjblue (commented):
Looking good! Left a few inline comments/ideas...

src/ember-app.js
suchitadoshi1987: do you mean 
``` let { app, appAnalytics } = shouldBuildApp
      ? await this.getNewApplicationInstance()
      : this.getAppInstanceInfo(this._applicationInstance);
```
should instead be
```
let { app, appAnalytics } = await this.getNewApplicationInstance(shouldBuildApp)
```
and the function would return the appropriate response

src/ember-app.js
melmerp: Rob,

Do we need all 3 of 'start', 'stop', and 'duration'? 

I'd prefer either (1) 'start' and 'stop' or (2) 'start' and 'duration'. I prefer the latter option more. The reason I like it is because: (1) it's smaller (which isn't a huge deal) and (2) whenever I see all 3 I do the subtraction to make sure 'duration' = 'stop' - 'start' because I always ask myself, why would they provide all 3?

That being said, this is not such a strong opinion and if you are strongly opinionated, I'll defer.

Thanks,
Mark

SergeAstapov:
@suchitadoshi1987 @rwjblue have you considered to instrument `result._finalize()` and `createShoebox()` as well?

src/ember-app.js
rwjblue: My reasoning here is _basically_ that "clock time duration" isn't enough information in all cases, but in many cases it is. For example, if we just provided duration it is actually quite difficult to map the results to see how many were interleaved/overlapping/whateverTheWordIs. In general, I feel like this is in a "why not" kinda scenario (where it costs us roughly nothing to do all three, and in fact we _basically_ have to know all three anyways in order to provide a duration). That being said, I definitely do **not** feel strongly about providing all 3. I think `start` and `duration` is probably enough info (and addresses my primary issue).

Let's call it "dealers choice", to the PR author go the spoils 😝

src/ember-app.js
rwjblue: Yes, IMHO we should have only a single function that is responsible for understanding:

* if we should build a new instance (based on `shouldBuildApp`)
* if we should build a new instance when the queue is empty and a new application instance is requested
* how to build the proper set of analytics output

rwjblue:
> @suchitadoshi1987 @rwjblue have you considered to instrument result._finalize() and createShoebox() as well?

Ya, we definitely can (and should!) add that. We should probably make an enumerated list over in https://github.com/ember-fastboot/fastboot/issues/107.

src/ember-app.js
suchitadoshi1987: lol okk.... i will unveil the suspense in my next commit  ;)

src/ember-app.js
suchitadoshi1987: agreed! 👍

src/ember-app.js
kratiahuja, line 217: Instead of doing this here, shouldn't these metrics come from the `queueObject` like `isItemPreBuilt`? It seems a bit of an odd API where we return whether the item was pre built or built on the fly but not the metrics of the time spent for that task. In other words, we are already getting what task was dequeued but not it's timing info. 

 If it isn't prebuilt, it can capture the time to create the sandbox. This way the generator function takes care of creation of the object, determining the state of reterival and metrics. If the intent is to get the time taken to built the sanbox then it should come from as part of the `dequeue` API IMO.

src/ember-app.js
kratiahuja, line 339: I think `visit` end time is when `finally` block is called in order capture the entire time. Is there specific reason why we are not capturing the remaining time to build the result and shoebox?

kratiahuja (changes requested):
I believe this also needs README updates as the public API of result.analytics is being changed: https://github.com/ember-fastboot/fastboot#result

src/ember-app.js
xg-wang: +1 to `performance.now()`, not only higher precision, more importantly it's using a monotonic clock therefore it is not subject to system clock skew or adjustments https://www.w3.org/TR/hr-time-2/#sec-monotonic-clock

src/ember-app.js
xg-wang: FWIW [`PerformanceEntry`](https://nodejs.org/api/perf_hooks.html#perf_hooks_class_performanceentry) defines `startTime` and `duration`

PR #277, opened 2020-10-23
Restore result finalization in resilient mode

bobisjan:
Previous version 2.x always finalised the result https://github.com/ember-fastboot/fastboot/blob/v2.0.3/src/ember-app.js#L363.

This PR restores that behaviour, which should unlock ember-fastboot/fastboot-express-middleware#70.


rwjblue (approved):
Thank you @bobisjan!

PR #282, opened 2020-10-30
Remove node version in Readme

xg-wang:
It's covered in package.json engines.
Also removes https://embadge.io badge which seems unavailable.


xg-wang:
hmmm
> AssertionError: expected '_emberMetalProperty_get.get(...).getURL is not a function' to equal 'App instance was forcefully destroyed in 5ms'

rwjblue (approved):
Thank you!

